PR #49, opened 2026-04-24
Update create-expo-module skill

behenate:
# Why

There is a lot of new changes in `create-expo-module`, we need to update the skill accordingly

# How

Add a reference file for `create-expo-module` 

# Test Plan

Tested by installing the skill locally and asking an agent to create an expo module. Looks like it its working good.

# Important

This skill includes information that will not be relevant until SDK-56 is published so we should correctly time when this change is published.

### DIFF
--- a/plugins/expo/skills/expo-module/SKILL.md
+++ b/plugins/expo/skills/expo-module/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: expo-module
-description: Guide for writing Expo native modules and views using the Expo Modules API (Swift, Kotlin, TypeScript). Covers module definition DSL, native views, shared objects, config plugins, lifecycle hooks, autolinking, and type system. Use when building or modifying native modules for Expo.
+description: Guide for creating and writing Expo native modules and views using the Expo Modules API (Swift, Kotlin, TypeScript). Covers module definition DSL, native views, shared objects, config plugins, lifecycle hooks, autolinking, and type system. Use when building or modifying native modules for Expo.
 version: 1.0.0
 license: MIT
 ---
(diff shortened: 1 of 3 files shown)


plugins/expo/skills/expo-module/SKILL.md
```diff
@@ -15,98 +15,66 @@ Complete reference for building native modules and views using the Expo Modules
...
+   - **Local module** for one app
+   - **Standalone module** for reuse, monorepos, or publishing
+2. Scaffold deliberately:
+   - pass an explicit slug or path
+   - choose `--platform` intentionally instead of relying on defaults
+   - add only the feature examples you actually want
+3. Replace generated example code with the real implementation.
+4. Remove unused generated files when needed:
+   - remove web bindings if the module is native-only
+   - remove view files if the module does not render UI
```
Kudo: could we select `--features` rather then remove unused files? given you've added the `--features` flag

so the workflow would be "determine what native features you need? a native module or a native component or so on"

plugins/expo/skills/expo-module/references/create-expo-module.md
```diff
@@ -0,0 +1,205 @@
...
+- the generated files do not match the existing module's shape
+
+If no features are detected or provided, the command creates a minimal scaffold for the new platform.
+
+## Environment Variables
+
+- `EXPO_BETA`: use the next template version
+- `EXPO_DEBUG`: enable debug logs
+- `EXPO_NO_TELEMETRY`: disable telemetry
+- `EXPO_NONINTERACTIVE`: force non-interactive mode
```
Kudo, line 205: since you mention EXPO_NONINTERACTIVE here. we may probably pick either `EXPO_NONINTERACTIVE` or `CI`

Kudo (approved):

sunilgentyala (approved):

Kudo (approved):

plugins/expo/skills/expo-module/references/create-expo-module.md
brentvatne: imo the env var should be called `EXPO_NON_INTERACTIVE` given that we write `non-interactive` so there is a separate between the words

behenate: We have already been using `EXPO_NONINTERACTIVE` in expo-module-scripts so I think it's not worth it to change it at this point

PR #58, opened 2026-05-12
Update UI skill names to match agent skill spec

mattr:
"Expo UI Jetpack Compose" and "Expo UI SwiftUI" names do not match the spec for agent skills (https://agentskills.io/specification#name-field).

Updates the name field in each skill to match the parent directory and the agent skill spec. This also brings these into line with the other Expo skills.


mattr:
Superseded by https://github.com/expo/skills/pull/80

PR #60, opened 2026-05-18
Add brownfield skill

gabrieldonadel:
This PR adds a skill that helps you with integrating Expo Brownfield into your existing native app.

### DIFF
--- a/plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
+++ b/plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
@@ -163,13 +163,13 @@ hermesEnabled=true
 
 ### `AndroidManifest.xml`
 
-Add the `INTERNET` permission to your main `AndroidManifest.xml`:
+Add the `INTERNET` permission to your main manifest at `app/src/main/AndroidManifest.xml`:
 
 ```xml
 <uses-permission android:name="android.permission.INTERNET" />
 ```
 
-In your **debug** `AndroidManifest.xml`, enable cleartext traffic so the app can talk to the local Metro bundler over HTTP:
+In the debug-variant manifest at `app/src/debug/AndroidManifest.xml`, enable cleartext traffic so the app can talk to the local Metro bundler over HTTP:
 
 ```xml
 <application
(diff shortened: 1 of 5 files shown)


Kudo:
wondering what's the test plan for this skill? could you share some steps and prompts when test the skill?

gabrieldonadel:
> wondering what's the test plan for this skill? could you share some steps and prompts when test the skill?

I tested this mostly with https://github.com/expo/expo-brownfield-examples

1. Double checked that the skill loaded without necessarily saying "brownfield" or "expo-brownfield".
  E.g. 
    1. "How do I add a React Native screen to my existing Kotlin app?"
    2. "We have a SwiftUI app and want to embed an Expo onboarding flow in one screen." 
 
2. Confirm Claude asks which approach to use based on team constraints
image

3. Check the implementation and confirm that the app runs as expected

plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
```diff
@@ -0,0 +1,468 @@
...
+}
+```
+
+Install pods:
+
+```sh
+cd ios && pod install
+```
+
+Open the generated `.xcworkspace` (not the `.xcodeproj`) from now on.
```
Kudo, line 344: [🚨 **critical**] The integrated iOS path jumps from `pod install` straight to AppDelegate, but the host Xcode project still needs the brownfield docs' project changes: set `ENABLE_USER_SCRIPT_SANDBOXING` to `No`, add the React Native bundle Run Script phase before `[CP] Embed Pods Frameworks`, and add `UIViewControllerBasedStatusBarAppearance=NO`. Without the bundle script, the release branch below looks for `main.jsbundle` but nothing has embedded it, so the React Native screen fails outside Metro.

plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
```diff
@@ -0,0 +1,468 @@
...
+
+### `AndroidManifest.xml`
+
+Add the `INTERNET` permission to your main `AndroidManifest.xml`:
+
+```xml
+<uses-permission android:name="android.permission.INTERNET" />
+```
+
+In your **debug** `AndroidManifest.xml`, enable cleartext traffic so the app can talk to the local Metro bundler over HTTP:
```
Kudo: maybe clear show the path like **src/debug/AndroidManifest.xml**

Kudo (approved):
looks good and we can first the skill first then iterates it

plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
gabrieldonadel: Ohh good catch!

gabrieldonadel: fixed

plugins/expo/skills/expo-brownfield/references/brownfield-integrated.md
gabrieldonadel: updated

PR #68, opened 2026-05-26
Add Convex integration skill

krystofwoldrich:
This is slightly generalized version of Expo Agent private skill used for setting up the Convex integration.

Open questions:
- Should we directly name alternatives in the skill description?
- How hard should agent recommend the integration based on the skill?
	- Instead of recommend maybe explicitly `Ask user if they specified alternative?` 
- Should we also add include specific skills for Convex's notification, auth or storage?

### DIFF
--- a/plugins/expo/skills/eas-integrations-convex/SKILL.md
+++ b/plugins/expo/skills/eas-integrations-convex/SKILL.md
@@ -1,9 +1,9 @@
 ---
-name: database
+name: eas-integrations-convex
 description: Use when an application has any type of data that should be saved or shared between users, when the user explicitly asks about a database, or when the user mentions a database tool such as Convex, Supabase, Postgres, Neon, PlanetScale, SQLite, or Prisma.
 ---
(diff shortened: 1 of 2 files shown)


plugins/expo/skills/eas-integrations-convex/SKILL.md
```diff
@@ -0,0 +1,47 @@
...
+bunx convex dev
+```
+
+5. Final step
+
+Inform user that convex is set up, running. To use the full Convex dashboard, user should join via invitation that was automatically sent to their email during this setup process.
+
+6. Latest Convex Docs
+
+For the latest Convex documentation, visit https://docs.convex.dev/quickstart/react-native.md
```
FiberJW, line 47: We might also want to link to this doc so the dev+agent can understand more about Convex than what the quickstart mentions: https://docs.convex.dev/understanding.md

FiberJW (approved):

PR #71, opened 2026-06-03
feat(expo): anonymous usage telemetry and skill feedback

zvadaadam:
## Summary

Adds **opt-in, anonymous usage telemetry** and a **feedback path** to the Expo skills plugin — so we can see which skills get used, and where they fall short. **Off by default: a fresh install sends nothing, creates nothing (not even a local id), until the user enables it.** Claude Code only; scripts are zero-dependency Node and run the same on macOS, Linux, and Windows.

## Signals

| Event | When it fires (only once enabled) | Wiring |
|-------|-----------------------------------|--------|
| `skill_invoked` | a plugin skill runs — `initiator: ai` (Skill tool) or `user` (`/slash` command) | plugin-level `hooks/hooks.json` (Claude Code) |
| `skill_feedback` | user-approved rating + short note | the `expo-skill-feedback` skill |

## Consent model

| | |
|---|---|
| Default | **Off.** No events, no feedback, no `~/.expo-skills/` dir, no install id. |
| Enable | tell the agent **"enable Expo skills telemetry"** → it runs `telemetry.cjs --on`; or `EXPO_SKILLS_TELEMETRY=1` |
| Disable | `--off` (default state), `EXPO_SKILLS_TELEMETRY=0`, or `DO_NOT_TRACK=1` — explicit-off always wins; CI never sends |
| Agent rules | the skill instructs the agent to answer status questions by running `--status`, and **never to enable telemetry on its own** — a refused feedback send asks the user once |

## Changes

- **Scripts** — zero-dependency Node (`.cjs`) in `plugins/expo/skills/expo-skill-feedback/scripts/`; the master gate is `telemetryActive()` in `telemetry_common.cjs`.
- **Cross-platform** — no shell wrappers; hooks and docs invoke `node` directly (works on cmd/PowerShell/sh). The hook self-detaches in Node (`spawn` + `unref`) so the POST never blocks the agent turn.
- **Hooks** — one plugin-level `hooks/hooks.json` (`PostToolUse[Skill]` + `UserPromptExpansion`); no per-skill hooks.
- **Feedback** — product skills carry **no footer**; feedback lives only in the `expo-skill-feedback` skill.
- **Docs** — README, plugin README, and AGENTS.md describe the opt-in model; version bumped to **1.6.0** across all three manifests.

## Privacy

- Anonymous: only the skill name, rating/note, platform (`os`/`arch`), and a **hash** of a random local install id — never PII, code, prompts, or file paths.
- The committed PostHog key is a public, **write-only** ingestion key (like a browser snippet key).
- **Proven with a network interceptor on a simulated fresh install:** telemetry off → zero outbound attempts and zero files created; enabled → exactly one POST to `us.i.posthog.com/i/v0/e/`.

## Testing

- ✅ 22-check offline harness: gate precedence, hook-payload resolution, plugin scoping, on/off marker, detached-child survival.
- ✅ Real headless Claude Code sessions (`claude -p --plugin-dir`): "enable expo skills telemetry" triggers the skill and writes the opt-in marker; status questions run `--status`; the hook fires but spawns nothing while off, and sends via the detached child once enabled.
- ✅ Live send returns HTTP 2xx; event shape identical to what was previously confirmed landing in PostHog project 375452.
- ✅ Repo CI checks pass: skill description/body limits, plugin version bump (1.5.0 → 1.6.0), `claude plugin validate`.


plugins/expo/skills/skill-feedback/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+
+```bash
+export EXPO_SKILLS_TELEMETRY=0   # or: false / off / no
+export DO_NOT_TRACK=1            # cross-tool standard, https://consoledonottrack.com
+```
+
+When disabled, no installation ID is created and nothing is sent. Uninstalling the plugin
+removes everything.
+
+## Configuring the PostHog project
```
davidmokos: Should this section be part of the skill?

plugins/expo/skills/skill-feedback/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+to commit. Override it per environment (e.g. a staging project) without editing the file:
+
+```bash
+export EXPO_SKILLS_POSTHOG_KEY="phc_your_project_key"
+```
+
+Only ever use a *project* key (`phc_...`) here. Never a *personal* API key (`phx_...`) —
+those are secret and used only for reading/querying (e.g. the PostHog MCP).
+
+## PostHog event shape
```
davidmokos: dtto

plugins/expo/skills/upgrading-expo/SKILL.md
```diff
@@ -132,3 +132,14 @@ Since SDK 55, users can opt-in to use Hermes engine v1 for improved runtime perf
...
 
 The new architecture is enabled by default, the app.json field `"newArchEnabled": true` is no longer needed as it's the default. Expo Go only supports the new architecture as of SDK +53.
+
+---
+
+## Expo Skill Feedback
+
+If this skill was useful, confusing, broken, or missing context, submit 1-3 safe sentences with the bundled feedback script:
+
+- **Claude Code:** `sh "${CLAUDE_SKILL_DIR}/../skill-feedback/scripts/run.sh" "${CLAUDE_SKILL_DIR}/../skill-feedback/scripts/skill-feedback.js" --skill upgrading-expo --rating idea --text "..."`
```
davidmokos: The UX here is not great imho. Can we create something like a `skill-feedback.sh` that will use the `run.sh` with the path to the js file as a param bundled in already?

So the command would be
```
sh "${CLAUDE_SKILL_DIR}/../skill-feedback/scripts/skill-feedback.sh" --skill upgrading-expo --rating idea --text "..."
```

plugins/expo/skills/upgrading-expo/SKILL.md
```diff
@@ -132,3 +132,14 @@ Since SDK 55, users can opt-in to use Hermes engine v1 for improved runtime perf
...
+- **Other agents (Codex, etc.):** run `skill-feedback/scripts/skill-feedback.js` (bundled in this plugin) with `node` or `bun`, same flags (the harness is auto-detected).
```
davidmokos: then this sh script can be reused here too.

plugins/expo/skills/skill-feedback/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+
+```bash
+sh "${CLAUDE_SKILL_DIR}/scripts/run.sh" \
+   "${CLAUDE_SKILL_DIR}/scripts/skill-feedback.js" \
+   --skill expo-deployment --rating useful \
+   --text "TestFlight steps were clear and worked first try."
+```
+
+**Other agents (Codex, etc.)** — run the same bundled script with `node` or `bun`,
+using its path on disk (the harness auto-detects; `--agent-harness` below is an
```
davidmokos: based on the code only codex can be auto-detected so this is misleading

AGENTS.md
```diff
@@ -252,3 +252,9 @@ codex plugin marketplace add expo/skills --ref main
 - Include a brief `README.md` for each plugin.
 - Keep references close to the skill that uses them.
 - Avoid broad rewrites when updating a skill; preserve the skill's existing scope and trigger intent.
+
+## Usage Telemetry & Feedback
+
+The `expo` plugin sends **anonymous** usage telemetry and collects skill feedback (details in the `skill-feedback` skill). Usage tracking is automatic via the plugin's `hooks/hooks.json` (no per-skill hooks). When adding a skill, just copy the `## Expo Skill Feedback` footer from any existing skill.
```
davidmokos: since it uses the plugin's hooks it *should* also be possible to add codex support. Codex supports plugin hooks

davidmokos (commented):
few comments inline

plugins/expo/skills/expo-skill-feedback/scripts/telemetry_common.cjs
```diff
@@ -0,0 +1,222 @@
...
+const path = require("path");
+
+const POSTHOG_HOST = "https://us.i.posthog.com";
+
+// PostHog project API key. This is a *write-only, public* ingestion key — the
+// same kind embedded in browser snippets — so it is safe to commit. Override
+// per environment with EXPO_SKILLS_POSTHOG_KEY (e.g. a staging project).
+// NOTE: never put a PostHog *personal* API key (phx_...) here — those are secret.
+const POSTHOG_PROJECT_API_KEY =
+  process.env.EXPO_SKILLS_POSTHOG_KEY || "phc_w8xRytdAAwkV3oExnuUozqH64PMzCmDLnyoChpPBcNXs";
```
Kudo, line 21: [🏗️ **design**] Because a real `phc_...` key is committed, `telemetryConfigured()` is true, so telemetry is **on by default** for every install — not "inert until a key is configured" as the PR description says. The only runtime disclosure is the one-time stderr line in `maybeShowFirstRunNotice()`, and it fires from a `--quiet` PostToolUse hook whose stderr Claude Code hides by default (transcript/debug only). So in practice the first send is silent.

The anonymization, opt-out, `DO_NOT_TRACK`, and CI-skip are all solid. The open question is consent: is silent on-by-default the intended model, or should first run surface something the user actually sees? Either way, please drop the "fully inert until a key is configured" line from the PR body, since a key is committed.

plugins/expo/skills/expo-skill-feedback/scripts/skill-event.cjs
```diff
@@ -0,0 +1,152 @@
...
+  const payload = eventPayload(skill, args, hookInput);
+
+  if (args.dryRun) {
+    console.log(JSON.stringify({ ...payload, api_key: "phc_..." }, null, 2));
+    return 0;
+  }
+
+  maybeShowFirstRunNotice();
+  try {
+    await sendToPosthog(payload, { userAgent: "expo-skills/skill-event", timeoutMs: 3000 });
```
Kudo, line 168: [💡 suggestion] The file header says this "never blocks," but `await`ing the POST keeps the hook process alive until PostHog responds or the 3s timeout fires — so each AI skill invocation can add up to 3s of latency to the turn (the hook waits for the process). For the quiet/auto path, consider detaching the send so telemetry stays off the critical path — e.g. background it in `run.sh` (`node "$script" "$@" &`) for the event scripts. Keep `await` in `skill-feedback.js`, since that path reports success to the user.

plugins/expo/hooks/hooks.json
```diff
@@ -0,0 +1,27 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Skill",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "sh \"${CLAUDE_PLUGIN_ROOT}/skills/skill-feedback/scripts/run.sh\" \"${CLAUDE_PLUGIN_ROOT}/skills/skill-feedback/scripts/skill-event.js\" --skill auto --initiator ai --plugin-root \"${CLAUDE_PLUGIN_ROOT}\" --quiet",
```
Kudo: note that `sh` doesn't support on windows

plugins/expo/skills/skill-feedback/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+  (openai/codex#21753) + enabling `plugin_hooks`; if it instead ships a *native* skill event,
+  the only addition is one matcher. Until then, `skill_feedback` is the Codex signal.
+- **Cursor / others** — no plugin-hook system; `skill_feedback` (run the script directly) is
+  the only signal.
+
+Manual `skill_feedback` works in every agent.
+
+## Turning it off
+
+Telemetry is anonymous and on by default. The reliable, launch-independent way to opt
```
Kudo: correct me if i am wrong. the skill-feedback skill is for agents without hook support to explicit evaluate the skill and share feedback.
in the case "on by default" or opt-out feels unclear to me if agent is going to explicitly send feedback

plugins/expo/skills/expo-skill-feedback/scripts/skill-feedback.cjs
```diff
@@ -0,0 +1,118 @@
...
+      $insert_id: "skill_feedback:" + crypto.createHash("sha256").update(insertSource).digest("hex").slice(0, 32),
+      source: SOURCE,
+      schema_version: SCHEMA_VERSION,
+      ...identityProperties,
+      agent_harness: agentHarness,
+      ...platformProps(),
+      skill,
+      rating: args.rating,
+      feedback_text: feedback,
+    },
```
Kudo, line 73: not sure if feasible: i think relevant prompts (or chat history) that triggers the skill would be helpful to further study how to improve the skill

Kudo (changes requested):
the main concern is windows which doesn't have `sh` support by default.

sunilgentyala (changes requested):
describe in detail

sunilgentyala (changes requested):
whats new in it?

plugins/expo/skills/skill-feedback/SKILL.md
zvadaadam: I simplified the skill-feedback and remove some of the unnecessary info.

plugins/expo/skills/skill-feedback/SKILL.md
zvadaadam: moved to /references

plugins/expo/skills/upgrading-expo/SKILL.md
zvadaadam: i dropped the other agents, we'll know only support claude code :/ codex hooks don't support the stuff we need yet...

plugins/expo/skills/skill-feedback/SKILL.md
zvadaadam: wording fixed

sunilgentyala (approved):

AGENTS.md
zvadaadam: codex currently doesn't support know what we need (maybe we can think of some workarounds in next PRs)

plugins/expo/skills/add-app-clip/SKILL.md
```diff
@@ -278,3 +278,11 @@ See [./references/native-module.md](./references/native-module.md) for the Swift
 ## References
 
 - ./references/native-module.md — Local Expo module to detect App Clip context and present the SKOverlay install prompt
+
+---
+
+## Expo Skill Feedback
```
davidmokos: I don't like that we're repeating the same paragraph for each skill here. Maybe it can be added via a CI script?

davidmokos (approved):
Lookin at the telemetry code I'm wondering whether to have `expo feedback` command in the CLI directly for that.

Kudo (approved):
thought i'm still not sure whether adding telemetry in the plugin would be a good idea. i don't want to block it. we could try that and see how it goes

plugins/expo/hooks/hooks.json
zvadaadam: Fixed in 772d40a — the `sh` wrappers are gone entirely. The hook (and the skill's documented commands) now invoke `node` directly, which runs the same under cmd, PowerShell, and POSIX shells, so Windows works without Git Bash. All the real logic was already zero-dependency Node; the shell layer was just launcher glue, now folded into `skill-event.cjs` itself.

plugins/expo/skills/expo-skill-feedback/scripts/skill-event.cjs
zvadaadam: Done in 772d40a — and in Node rather than `run.sh &` so it stays cross-platform: the hook process reads the payload, resolves the skill, runs the cheap local gates, then re-launches itself detached (`spawn(..., {detached: true, stdio: 'ignore', windowsHide: true}).unref()`) to do the POST, returning immediately — the turn no longer waits on the network. `skill-feedback.cjs` keeps the foreground `await` since it reports success to the user, exactly as you suggested. This also let us delete all three `.sh` wrappers (−67 lines of shell).

plugins/expo/skills/expo-skill-feedback/scripts/telemetry_common.cjs
zvadaadam: You were right to push on this — we ended up going further than a louder disclosure: telemetry is now **opt-in and off by default** (061f807). A fresh install sends nothing and creates nothing (not even the local install id) until the user enables it — `telemetry.cjs --on`, `EXPO_SKILLS_TELEMETRY=1`, or just telling the agent "enable Expo skills telemetry"; `DO_NOT_TRACK`/explicit-off/CI always win. Verified with a network interceptor on a simulated fresh install: zero outbound attempts while off, exactly one POST to PostHog once enabled. The stale "fully inert until a key is configured" line is gone — the PR description now describes the opt-in model.

plugins/expo/skills/skill-feedback/SKILL.md
zvadaadam: This got resolved by the opt-in pivot (061f807): there is no on-by-default anything anymore. Both signals — the automatic `skill_invoked` hook event and the explicit `skill_feedback` — sit behind the same opt-in gate, so even an agent explicitly sending feedback sends nothing unless the user enabled telemetry; when it's off, the script refuses and the skill tells the agent to ask the user once (and never to enable it on its own). We also dropped the feedback footers from all product skills — feedback now lives only in the `expo-skill-feedback` skill.

plugins/expo/skills/expo-skill-feedback/scripts/skill-feedback.cjs
zvadaadam: Considered it, but decided to keep prompts/chat history out on purpose — it's the bright line of the privacy model (the skill promises never to send prompts, code, or paths), and raw transcripts can't be reliably scrubbed of secrets/PII. The substitute is the `--text` note: the agent writes 1–3 scrubbed sentences of qualitative context (what was confusing or missing), which keeps the 'why' for improving the skill without shipping anyone's conversation. If we need richer signal later, I'd add structured opt-in fields rather than raw history.

plugins/expo/skills/expo-skill-feedback/scripts/telemetry_common.cjs
```diff
@@ -0,0 +1,171 @@
...
+function telemetryActive() {
+  const env = telemetryEnvSignal();
+  if (env === "off") return false;
+  if (isCI()) return false;
+  if (env === "on") return true;
+  try { if (fs.existsSync(OPT_IN_PATH)) return true; } catch {}
+  return DEFAULT_ON;
+}
+
+// A real key ships in this file, so telemetry is ON by default. This guard only
```
Kudo: [💡 suggestion] Stale comment from the opt-out iteration — it says telemetry is ON by default, contradicting `DEFAULT_ON = false` above and the whole PR. What's true by default is that this *check* passes; anyone auditing the privacy behavior shouldn't trip over this.

```suggestion
// A real key ships in this file, so this check passes by default. This guard only
```

plugins/expo/skills/expo-skill-feedback/scripts/skill-event.cjs
```diff
@@ -0,0 +1,177 @@
...
+//   - Claude Code Skill tool:        tool_input.skill        (e.g. "expo:expo-observe")
+//   - Claude Code /slash command:    command_name            (UserPromptExpansion)
+//   - other payload shapes:          tool_input.skill_name, top-level skill / skill_name
+// Plugin skills are namespaced (e.g. "expo:expo-observe") — keep the final segment.
+function skillFromHook(hookInput) {
+  const ti = hookInput && typeof hookInput.tool_input === "object" && hookInput.tool_input ? hookInput.tool_input : {};
+  const raw = String(
+    ti.skill || ti.skill_name || hookInput.command_name || hookInput.skill || hookInput.skill_name || ""
+  ).trim().replace(/^\//, ""); // tolerate a leading "/" from slash-command payloads
+  return raw.includes(":") ? raw.slice(raw.lastIndexOf(":") + 1) : raw;
```
Kudo: [💡 suggestion] The namespace is stripped before the scoping check, so another plugin's skill with a colliding name is counted as ours: a payload of `tool_input.skill = "foo:expo-ui"` resolves to `expo-ui` and passes `skillBelongsToPlugin()` (verified with a crafted hook payload). When a namespace is present, require it to be `expo`:

```suggestion
  if (!raw.includes(":")) return raw;
  const sep = raw.lastIndexOf(":");
  return raw.slice(0, sep) === "expo" ? raw.slice(sep + 1) : "";
```

If harnesses ever send a differently-qualified namespace this would drop real events, so feel free to loosen — but the current code over-counts, and the comment above claiming downstream scoping "keeps this safe" doesn't hold for name collisions.

plugins/expo/skills/expo-skill-feedback/scripts/skill-feedback.cjs
```diff
@@ -0,0 +1,118 @@
...
+  };
+}
+
+async function main(argv) {
+  const args = parseArgs(argv);
+
+  if (!args.dryRun && !telemetryActive()) {
+    console.error("skill-feedback: telemetry is off (opt-in, off by default); nothing sent. Enable with `telemetry.cjs --on` or EXPO_SKILLS_TELEMETRY=1.");
```
Kudo:
```suggestion
    console.debug("skill-feedback: telemetry is off (opt-in, off by default); nothing sent. Enable with `telemetry.cjs --on` or EXPO_SKILLS_TELEMETRY=1.");
```

since it's off by default. error log level feels too strong

Kudo (approved):

PR #79, opened 2026-06-09
add plugin version bump check

davidmokos:
## Summary
- add a dependency-free Node script that checks Expo plugin version bumps against origin/main
- require Claude, Codex, and Cursor plugin manifests to stay in sync and increase when versioned plugin files change
- add a GitHub Actions workflow that runs on pull_request and merge_group, writes a summary, and posts a sticky PR comment when possible

## Validation
- node --check scripts/check-plugin-version-bump.mjs
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/plugin-version-bump.yml')"\n- node scripts/check-plugin-version-bump.mjs origin/main\n- simulated missing manifest, false-positive path, valid bump, and prerelease bump cases in disposable worktrees

### DIFF
--- a/.github/workflows/plugin-version-bump.yml
+++ b/.github/workflows/plugin-version-bump.yml
@@ -15,7 +15,6 @@ concurrency:
 permissions:
   contents: read
   issues: write
-  pull-requests: read
 
 jobs:
   check:
@@ -28,12 +27,15 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+
       - name: Check plugin version bump
         id: version-check
         continue-on-error: true
         env:
           VERSION_CHECK_SUMMARY_PATH: .context/plugin-version-check-summary.md
-        run: node scripts/check-plugin-version-bump.mjs origin/main
+        run: bun scripts/check-plugin-version-bump.ts origin/main
 
       - name: Add workflow summary
         if: always() && hashFiles('.context/plugin-version-check-summary.md') != ''
--- a/.github/workflows/plugin-version-bump.yml
+++ b/.github/workflows/plugin-version-bump.yml
@@ -1,0 +1,84 @@
+name: plugin version bump
+
+on:
+  pull_request:
+    branches:
+      - main
+  merge_group:
+    branches:
+      - main
+
+concurrency:
+  group: plugin-version-bump-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: read
+
+jobs:
+  check:
+    name: check Expo plugin versions
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/plugin-version-bump.yml
```diff
@@ -0,0 +1,84 @@
...
+        continue-on-error: true
+        env:
+          VERSION_CHECK_SUMMARY_PATH: .context/plugin-version-check-summary.md
+        run: node scripts/check-plugin-version-bump.mjs origin/main
+
+      - name: Add workflow summary
+        if: always() && hashFiles('.context/plugin-version-check-summary.md') != ''
+        run: cat .context/plugin-version-check-summary.md >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Comment on pull request
```
Kudo, line 44: [💡 suggestion] Heads up on the fork-PR case: for `pull_request` events from forks, GitHub hands the workflow a read-only `GITHUB_TOKEN` regardless of the `issues: write` block, so `issues.createComment` returns 403 and the sticky comment never posts. `continue-on-error` keeps it non-fatal, and the job summary still renders — but since most PRs to a public repo come from forks, in practice external contributors will only ever see the summary (under the Actions tab), never the inline comment.

That may be exactly the intended "when possible" behavior. Just confirming it's acceptable that fork PRs get summary-only feedback, vs. a `workflow_run`-triggered comment job (the GitHub-recommended safe pattern for commenting on fork PRs). I'd avoid `pull_request_target` here since it would run the checked-out PR script with a write token.

.github/workflows/plugin-version-bump.yml
```diff
@@ -0,0 +1,84 @@
...
+      - main
+
+concurrency:
+  group: plugin-version-bump-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: read
```
Kudo:
```suggestion
  contents: read
  issues: write
```

would be good to minimize the permissions as much as possible. not sure if pull_requests permission is used?

scripts/check-plugin-version-bump.mjs
```diff
@@ -0,0 +1,254 @@
...
+
+function readBaseJson(path) {
+  try {
+    return JSON.parse(runGit(["show", `${baseRef}:${path}`]));
+  } catch {
+    return null;
+  }
+}
+
+function parseSemver(version) {
```
Kudo: we can try to use bun which has semver as builtin https://bun.com/docs/runtime/semver. then we don't need to implement semver here and reduce maintaince cost

we can also write the script in typescript directly

.github/workflows/plugin-version-bump.yml
Kudo: explaining clearly: pr comments for pr from external fork won't work. that's fine currently because `pull_request_target` isn't a good idea

Kudo (changes requested):
recommend to use bun and save custom semver implementation. that's minor one and up to you to make to call. if you go for current design, just re-request review for me

Kudo (approved):
thanks for the following up

PR #81, opened 2026-06-11
Dedupe EAS Workflow syntax out of the expo-deployment skill

Kudo:
# Why

close ENG-21677

# How

- dedupe workflow syntax from expo-deployment skill and recommend to use expo-cicd-workflows skill
- fine tune skill description
- add TOC for ios-app-store.md and app-store-metadata.md because they are large markdowns

### DIFF
--- a/plugins/expo/skills/expo-deployment/references/app-store-metadata.md
+++ b/plugins/expo/skills/expo-deployment/references/app-store-metadata.md
@@ -2,6 +2,24 @@
 
 Manage App Store metadata and optimize for ASO using EAS Metadata.
 
+## Contents
+
+- [What is EAS Metadata?](#what-is-eas-metadata)
+- [Getting Started](#getting-started)
+- [Configuration File](#configuration-file)
+- [App Store Optimization (ASO)](#app-store-optimization-aso)
+- [Categories](#categories)
+- [Localization](#localization)
+- [Dynamic Configuration](#dynamic-configuration)
+- [Age Rating (Advisory)](#age-rating-advisory)
+- [Release Strategy](#release-strategy)
+- [Review Information](#review-information)
+- [ASO Checklist](#aso-checklist)
+- [VS Code Integration](#vs-code-integration)
+- [Common Issues](#common-issues)
+- [CI/CD Integration](#cicd-integration)
+- [Tips](#tips)
+
 ## What is EAS Metadata?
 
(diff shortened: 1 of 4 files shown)


plugins/expo/skills/expo-deployment/references/app-store-metadata.md
```diff
@@ -2,6 +2,24 @@
 
 Manage App Store metadata and optimize for ASO using EAS Metadata.
 
+## Contents
```
davidmokos, line 5: Is the contents section useful or just wasted tokens?

plugins/expo/skills/expo-deployment/references/ios-app-store.md
```diff
@@ -1,5 +1,21 @@
 # Submitting to iOS App Store
 
+## Contents
```
davidmokos, line 3: dtto

plugins/expo/skills/expo-deployment/references/ios-app-store.md
```diff
@@ -1,5 +1,21 @@
 # Submitting to iOS App Store
```
davidmokos, line 1: Do we need this file at all when we have launch.dev? I'd rather have a short skill that tells the agent to go to launch.dev with the prefilled parameters. No need to do all this manually

davidmokos (commented):
This is unrelated to your changes but should we use launch.dev for pushing to app store instead of doing it manually?

plugins/expo/skills/expo-deployment/references/app-store-metadata.md
Kudo: i tried to use skill-creator and it's from the recommendation

> For large reference files (>300 lines), include a table of contents

https://github.com/anthropics/skills/blob/57546260929473d4e0d1c1bb75297be2fdfa1949/skills/skill-creator/SKILL.md?plain=1#L98

open to remove it if that's not useful. wdyt?

plugins/expo/skills/expo-deployment/references/ios-app-store.md
Kudo: good point maybe we could have a separate pr to follow it up. we should also verify how good agent to do non-interactive through launch.
added a task https://linear.app/expo/issue/ENG-21870/update-ios-app-store-skill-references-with-launch

davidmokos (approved):

PR #87, opened 2026-06-22
add expo-skill-eval

Kudo:
# Why

add evals for our skills

# How

inspired by the skill-creator that we can use a skill to run evals. 

before running the eval, it will ask some questions

- which skill to run the eval
- what to verify: screenshots, lint results, or some custom criteria
- what prompts to drive the evals, having some default e.g. build settings screen for expo-ui skill. you can also upload a screenshot to ask agent to generate code by the screenshot
- which sdk version for the projects (will fetch the latest sdk version as default)
- whether to test on expo-go or development builds
- which platforms to test (android, ios, web)
- for eval result, keep it on local or upload to claude shareable artifact

- confirm question to run subagents on `--dangerously-skip-permissions` mode. otherwise subagents will ask many questions
- confirm question to make sure the installed expo plugin is disabled. otherwise people may use the skills from marketplaces rather than the local developed

# Test Plan

multiple runs on sonnet 4.6. some shared results that required expo org claude accounts to access

- https://claude.ai/code/artifact/4dddcbc2-20af-44e5-aa22-296ca6c63c4f
- https://claude.ai/code/artifact/ce63b635-9ff6-4062-b408-fd335fa10a4b


.claude/skills/expo-skill-eval/scripts/__pycache__/generate_viewer.cpython-312.pyc
davidmokos, line 1: Should we add __pycache__ to .gitignore?

davidmokos (approved):
This is super helpful for us. The questions is whether it's helpful for expo users and should be shipped with the rest of the skills?

Kudo:
> The questions is whether it's helpful for expo users and should be shipped with the rest of the skills?

​that's not a scope currently. the skill is based on many assumptions and only works on macos host. if people really want to have the skill, they could copy the entire directory to their $HOME/.claude/skills.

zvadaadam (approved):
let's merge it so we can start experimenting with it <3

I had some problems with expo go in this version but might be problem with my setup. I don't feel it's a blocker now.

PR #88, opened 2026-06-24
Add eas-simulator skill

gwdp:
### Why

EAS provides cloud simulators. This skill makes it easy for agents to open in-development apps on them.

### What

New skill at `plugins/expo/skills/eas-simulator/` — SKILL.md and three reference files. Covers the session lifecycle, three ways to get an app onto the sim, driving it (press, screenshot, etc.), and cleanup.

### Test plan

Ran all three modes on real sessions from Cursor on a Mac. Opus 4.8, Composer 2.5, and Codex 5.3 all picked up and followed the skill correctly.

<details>
<summary>Mode A — local release build</summary>

Built locally, uploaded to the cloud sim, app rendered.

screenshot

[Transcript](https://github.com/user-attachments/files/29320696/screenshot.md)

</details>

<details>
<summary>Mode B — EAS build (iOS)</summary>

EAS built it, VM downloaded the artifact (no credentials), app rendered.

eas_build

[Transcript](https://github.com/user-attachments/files/29322148/eas_build.md)

</details>

<details>
<summary>Mode B — EAS build (Android)</summary>

EAS built the APK, VM downloaded it, rendered on the remote Android emulator. No browser preview — Android is screenshot-only.

android

[Transcript](https://github.com/user-attachments/files/29360330/android.md)

</details>

<details>
<summary>Mode C — dev build + tunnel (live edit)</summary>

Edited a file locally, it hot-reloaded on the remote sim with no rebuild.

before
after hot reload

[Transcript](https://github.com/user-attachments/files/29319852/cloud-sim.md)
[Transcript Linux](https://github.com/user-attachments/files/29372414/linux-mode-c.md)

</details>

<details>
<summary>argent — editor MCP link</summary>

Started an argent session (`--type argent`), ran `argent init --yes` to register the Argent MCP in Cursor, wired `ARGENT_TOOLS_URL` + `ARGENT_AUTH_TOKEN` as env vars in `.cursor/mcp.json` (higher precedence than `argent link`, works in sandboxed shells), and reloaded the editor. The remote sim was controllable from chat via Argent's MCP tools.

[Transcript](https://github.com/user-attachments/files/29360344/argent-good.md)

</details>

<details>
<summary>Trigger accuracy — 6/6 (100%)</summary>

| # | Prompt | Result |
|---|--------|--------|
| 1 | Run my Expo app on a cloud iOS simulator and screenshot the home screen | ✅ |
| 2 | I'm on Linux with no Mac available — use EAS to open my app on a remote simulator | ✅ |
| 3 | Start an EAS simulator session, install my app, and take a screenshot | ✅ |
| 4 | Use eas simulator to run my expo app on iOS and capture a screenshot of it | ✅ |
| 5 | I want to make live code changes and see them hot-reload on a cloud simulator without a Mac | ✅ |
| 6 | Open my app on a remote Android simulator via EAS and verify it renders correctly | ✅ |

</details>

Still to do:
- Check whether this belongs in public skills while the feature is experimental


plugins/expo/skills/eas-simulator/references/run-your-app.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Mode B — EAS build (the VM downloads it; no credentials)
+
+**Explicit-only** (see the SKILL.md mode picker): a *static* EAS artifact for CI/sharing, or when the user names an existing EAS build. For no-Mac **live** iteration use Mode C with an EAS dev-client build (see Mode C below), not this. **Simulator builds are unsigned, so EAS asks for no credentials.**
+
+⚠️ **Order matters:** build FIRST, `start` the session LAST. The build takes ~15-20 min and a session left idle that long times out (`ERR_NGROK_3200`) — don't `start` until you have the artifact URL.
+
+```bash
+# 1. Add a simulator build profile to eas.json
+cat > eas.json <<'JSON'
```
davidmokos: This should not overwrite eas.json. Agents often run skill command blocks as written, so this can delete existing build or submit profiles.
A better way would be to instruct the agent to merge/add a sim profile while preserving existing cli, build, and submit config

davidmokos (commented):
Review comments based on Anthropic and OpenAI skill-creator guidance.

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,145 @@
+---
+name: eas-simulator
+description: "Run and control a user's app on a remote iOS/Android simulator hosted on EAS cloud. Always read before executing any `eas simulator:*` commands — it has the current syntax for this experimental API. Use whenever the user needs a simulator they can't run locally — 'run my app on a cloud simulator', 'use eas simulator to run/install/screenshot my app', 'I'm on Linux/Cursor and need an iOS device', 'no sim on this box / headless CI', 'let an agent click through my app and screenshot it', 'test my dev build on a remote sim with live reload', 'stream a sim's screen to my browser' — even when they don't say 'EAS Simulator' or 'cloud'. Covers `eas simulator:*` plus installing a build and a dev-build live-edit loop. On a host WITHOUT a local simulator (Linux, CI, cloud sandbox) it's the default — just use it; on macOS, where a local sim exists, do NOT auto-trigger for a plain 'run on the simulator' — use it only for a cloud/remote/shareable sim, an iOS version they lack, or an agent-driven session. NOT for local sims (expo run:ios, Xcode, Android Studio), EAS Build/Update, web preview, or physical devices."
```
davidmokos: This description is 1,115 chars, we should fit into 1,024 chars. Please trim it. Keep the main trigger cases here and move examples or exclusions into the body.

plugins/expo/skills/eas-simulator/references/run-your-app.md
```diff
@@ -0,0 +1,158 @@
...
+
+---
+
+## Mode C — Local dev build + tunnel (live edits via Fast Refresh)
+
+This is the agentic edit-and-see loop: a **dev (Debug) build** loads JS from your local **Metro** over **tunnel v2**, so code edits appear on the remote sim via Fast Refresh. It has the most steps — each is necessary.
+
+⚠️ **Don't install a release build as a "quick interim" and screenshot it** — that interim shows stale, build-time code (the "outdated screenshot" trap). Go straight to the dev build + Metro; screenshot only after the dev client is connected to Metro.
+
+**No local Mac toolchain?** (the common cloud/Linux case) Build the dev client on **EAS** instead of step 1: add a profile with `{"developmentClient": true, "ios": {"simulator": true}}`, run `npx --yes eas-cli@latest build --platform ios --profile dev-sim`, and in step 3 use `install-from-source <artifact-url>` (VM downloads it) instead of the local `install`. Steps 2 and 4–7 (Metro tunnel + connect) are identical — your edited JS still loads live from your local Metro.
```
davidmokos: I'm confused a little here - Mode C is the default and the main point of cloud simulator is that you can run on non-macOS machine. However, the EAS dev-client path is only a short paragraph even though it's the most important part. Can we expand it a little?

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+version: 1.0.0
+license: MIT
+allowed-tools: "Bash(npx *), Bash(eas *), Bash(expo *), Bash(xcodebuild*), Bash(pod*)"
```
Kudo: can we scope the tools to smaller group? especially the `npx *` feels dangerous

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+
+# EAS Simulator
+
+EAS Simulator runs a remote iOS simulator or Android emulator on EAS infrastructure that you drive from your machine — from the CLI, from an AI agent (via `agent-device`), and from a browser preview. It's the unlock for **environments that can't run a simulator locally** (Linux boxes, cloud/background agents like Cursor Cloud), and for letting an agent *verify* a change on a real device instead of only reasoning about code.
+
+The `simulator:*` commands are **experimental and hidden**, and need a recent eas-cli (≥ 20.3.0 as of writing) — which is why this skill runs everything via `npx --yes eas-cli@latest`. Flags and verbs may change; if a command fails, **`<cmd> --help` is authoritative.**
+
+## When to use
+
+The frontmatter `description` carries the trigger phrases. In short: use this to get a user's app onto a **cloud** simulator and interact with it — especially from a Mac-less or cloud/sandbox agent. **Not** for local sims (`expo run:ios`, Xcode, Android Studio), store builds/signing (that's EAS Build), or physical devices. For the macOS case, see *Cloud vs local* next.
```
Kudo, line 17: i suppose agents will load frontmatter and description before this section?

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Cloud vs local: decide this first
+
+- **Non-macOS** (Linux / CI / cloud sandbox like Cursor Cloud, detect via `uname -s` ≠ `Darwin`): the only way to get a sim — **just proceed.**
+- **macOS:** local sims exist and a cloud session costs money + latency, so **ask first** ("a remote cloud sim — to share a live preview, offload, or test an iOS version you lack — or just run locally?") unless the user explicitly said cloud/remote/shareable.
```
Kudo, line 22: can we run some local script to determine whether local simulator/emulator is available?
e.g. checking `simctl` or xcodebuild available?

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+
+- **Run every `eas` command via `npx --yes eas-cli@latest …`** — guarantees a CLI new enough to have `simulator:*` (a global `eas` is often too old), and `--yes` skips npx's prompt. (Bare `eas` is fine if `eas --version` is current.)
+- **Authenticated.** Interactive machine → `npx --yes eas-cli@latest login`. **Cloud sandbox / CI / headless agent has no browser login — set `EXPO_TOKEN`** (expo.dev → Account → Access Tokens) in the env instead. Verify either way with `npx --yes eas-cli@latest whoami`.
+- Run from an Expo **project directory.** A fresh app needs one-time setup: `npx --yes eas-cli@latest init` to create/link the project (when there's no `projectId`), and **set `ios.bundleIdentifier`** in app config if it's missing — a fresh `create-expo-app` often has none, and `prebuild`/`eas build` need it (they prompt or fail without it; e.g. `dev.<owner>.<slug>`). Read current config with `npx expo config --json` (it may live in `app.config.js`). The first Mode-C run is slow (native build); later runs reuse it.
+- A controller to drive the device. This skill uses **agent-device** (open source, MIT), run on demand via `npx agent-device@latest` — nothing globally installed.
```
Kudo: should we mention argent or other alternatives here in Prerequisites?

Kudo (commented):
would like to know further test plans:
- if we ask agent to use argent, does it work well? (because the skill / references hardcoded agent-device in multiple places)
- if we ask agent to run android workflow, does ti work well? (because the skill / references hardcoded ios in multiple places)

plugins/expo/skills/eas-simulator/references/run-your-app.md
gwdp: Thank you for bringing that up. 🙏 I have updated the instructions to reflect a very specific change instead

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: trimmed without significant loss. Will also try to follow up with a simple CI check, wdty?

plugins/expo/skills/eas-simulator/references/run-your-app.md
gwdp: It was written on top of the Mac sequence since the Metro/connect steps are identical afaik. Expanded it into a full standalone block with the dev-sim profile, EAS build command and install-from-source step. Validated end-to-end from Docker on Linux (Mode C) — transcript in the PR 🎉

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: Thank you for calling this out 🙏 Scoped to `eas-cli@*`, `agent-device@*,` and `expo`, covers everything the skill actually runs (for now)

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: 100%, dropped a snippet right there

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: Done, and went a bit further. Rewrote `controllers.md` after testing end-to-end. Transcript added in the PR.

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: That's my assumption — "When to use" is intentional expanded guidance for once the full body is in context, but happy to be corrected if the loading order works differently

gwdp:
> would like to know further test plans:
> 
> * if we ask agent to use argent, does it work well? (because the skill / references hardcoded agent-device in multiple places)
> * if we ask agent to run android workflow, does ti work well? (because the skill / references hardcoded ios in multiple places)

Thanks for the review @Kudo and @davidmokos 🙏 Addressed all the feedback — replies inline. Also added test plans for argent (editor MCP link) and Android, both worked quite well, although argent received a little refinement to improve experience. Let me know how you guys feel about it!

plugins/expo/skills/eas-simulator/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Cloud vs local: decide this first
+
+- **Non-macOS** (Linux / CI / cloud sandbox like Cursor Cloud, detect via `uname -s` ≠ `Darwin`): the only way to get a sim — **just proceed.**
+- **macOS:** local sims exist and a cloud session costs money + latency, so **ask first** ("a remote cloud sim — to share a live preview, offload, or test an iOS version you lack — or just run locally?") unless the user explicitly said cloud/remote/shareable.
+- Always honor an explicit choice; for "run it locally" hand off to `expo run:ios` / Xcode.
+
+```bash
+# Programmatic detection — run this to decide before doing anything else:
+if [ "$(uname -s)" != "Darwin" ] || ! command -v simctl &>/dev/null; then
```
Kudo: [🏗️ **design**] `command -v simctl` always fails on macOS — `simctl` isn't on `PATH`, it's reached via `xcrun simctl` (it lives inside Xcode at `.../Developer/usr/bin/simctl`). So `! command -v simctl` is always true and this branch always prints "proceed with EAS Simulator", even on a Mac with simulators. That defeats the "on macOS, ask first" rule two lines above and can spin up a paid cloud session without asking. Verified `command -v simctl` exits 1 on a Mac with Xcode installed.

```suggestion
if [ "$(uname -s)" != "Darwin" ] || ! xcrun --find simctl &>/dev/null; then
```

README.md
```diff
@@ -120,6 +120,7 @@ Agents choose the right skill from the task context and each skill's description
 | `expo-cicd-workflows` | EAS Workflow YAML files and CI/CD automation. |
 | `expo-observe` | EAS Observe setup and launch, route, event, and version metrics. |
 | `eas-update-insights` | EAS Update health, crash rates, launch counts, payload size, and rollout gates. |
+| `eas-simulator` | Run and drive your app on a remote iOS/Android simulator on EAS cloud — from the CLI or an agent, with a live browser preview. |
```
Kudo: [🐧 nit] "with a live browser preview" overclaims for Android — the skill itself notes the preview is iOS-only (Android is screenshot/agent-only). `plugins/expo/README.md` already drops the claim. Qualify it here:

```suggestion
| `eas-simulator` | Run and drive your app on a remote iOS/Android simulator on EAS cloud — from the CLI or an agent, with a live browser preview (iOS). |
```

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,71 @@
...
+| `snapshot -i` | Interactive accessibility tree → `@e1`-style refs. iOS snapshots can be slow (tens of seconds). |
+| `press <ref\|selector> [x y]` | **Tap.** The verb is `press`, NOT `tap`. Selector form e.g. `press 'label="Open"'`. |
+| `fill <ref> "text"` | Type into a field |
+| `screenshot <path>` | Capture to a local PNG (downloaded from the daemon) |
+| `gesture <pan\|fling\|swipe\|pinch\|rotate\|transform>` / `swipe` / `scroll` / `longpress` / `type` / `back` / `home` | Standard interactions — `gesture` needs a kind subcommand |
+| `metro prepare (--public-base-url <url> \| --proxy-base-url <url>)` / `metro reload` | Wire a dev client to Metro / reload (Mode C). `--proxy-base-url` = optional bridge origin for remote Metro access — **not exercised by this skill's flow** (run-your-app.md connects via the public tunnel URL). |
+| `logs` / `record` / `network` / `perf` | Evidence capture |
```
Kudo: nit: i feel that we should not doc agent-device in our skill. does agent-device have some doc that we can ask agents to fetch?

or even just run `agent-device help`?

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,71 @@
...
+
+Run `npx --yes eas-cli@latest simulator:exec npx agent-device@latest --help` (and `<verb> --help`) for the authoritative, current set — the CLI help is the source of truth.
+
+**Proven in this skill's flows:** `apps`, `install`, `install-from-source`, `open`, `snapshot -i`, `press`, `fill`, `screenshot`. The rest (`gesture`/`scroll`/`longpress`/`logs`/`record`/`network`/`perf`/`metro`) are real in the CLI but not exercised here — confirm shape via `<verb> --help` before relying on them.
+
+## argent (alternative)
+
+`npx --yes eas-cli@latest simulator:start --type argent` provisions an argent remote session. The connection config it returns is different (`ARGENT_TOOLS_URL` / `ARGENT_AUTH_TOKEN`).
+
+**argent is a driver, not an installer.** Its `reinstall-app` tool runs `xcrun simctl install` on the remote VM with a literal local path — it does not upload a binary or accept a URL. **Always use agent-device (or `install-from-source`) for the install step**, then use argent to drive the already-installed app.
```
Kudo: > Always use agent-device (or `install-from-source`) for the install step**, then use argent to drive the already-installed app.

that's....interesting 🤔 i think our tooling (eas-cli / workflow) should run `xcrun simctl install` than asking agent/user to use agent-device just for installing an app

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,71 @@
...
+      "env": {
+        "ARGENT_TOOLS_URL": "<ARGENT_TOOLS_URL from simulator:start>",
+        "ARGENT_AUTH_TOKEN": "<ARGENT_AUTH_TOKEN from simulator:start>"
+      }
+    }
+  }
+}
+```
+
+`.cursor/mcp.json` now carries a session token — **add it to `.gitignore`**. Then reload Cursor (Cmd+Shift+P → "Reload Window") and Argent's MCP tools drive the remote sim from chat.
```
Kudo: should we mention other agents like claude code / codex here?

ideally we should write a script to update the `mcp.json` ARGENT_TOOLS_URL / ARGENT_AUTH_TOKEN for people (could put the script in the skill). the script could detect different agents and also people install mcp under user root or project root

Kudo (changes requested):
i have a mixed feeling for the current work. but i don't want to block much. if you want to ship it first and iterate later. please just re-request review for me. (i use graphite dashboard. please don't feel hesitant about the "request changes" / "re-request review" workflow)

the major concern is
- some step-by-step in markdowns could potentially use scripts in the skill to make it more deterministic. for example, the argent mcp setup.
- i feel we are mixing "how-to build" concept in this skill. ideally we should keep this skill smaller and link to other skill. currently the differece between mode a/b/c are mostly how the app built. we could try to focus single stuff - "given an app/ipa/apk, use eas simulator" to run it. how to generate app/ipa/apk is not in the skill scope.
- imo, mode b is the happy path for me. even though the skill said it's rarely used. if people can build ios app locally, they should mostly have local simulators available. using eas build or eas workflow to build the app, for people running on linux/windows machine to test ios app, is something we are targeting first. maybe we can simplify the mode matrix a little bit.

plugins/expo/skills/eas-simulator/SKILL.md
davidmokos: that would be great!

README.md
davidmokos: we should keep the consistent naming everywhere: ios simulator vs android emulator

plugins/expo/skills/eas-simulator/references/controllers.md
davidmokos: agent device skills are installed when you install agent-device: https://github.com/callstackincubator/agent-device

(side note: did we test the full process including the agent-device install on windows?)

davidmokos: also, we should probably commit to a single library: `argent` or `agent-device`, instead of recommending both

davidmokos (approved):
Don't want to block this. I think we need to test it properly in the wild and make changes when we see the agents struggle. I'd encourage everyone in expo to dogfood this skill and the simulator service after we merge it

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: Thanks for catching that, updated

README.md
gwdp: Good call 🙇‍♂️ Updated with both suggestions!

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: > nit: i feel that we should not doc agent-device in our skill. does agent-device have some doc that we can ask agents to fetch?
> 
> or even just run `agent-device help`?

Agreed that `--help` is better and decouples us from maintaining a verb list in the skill. Updated. I originally added the table after seeing models struggle with agent-device in dogfooding sessions. Will do a bit more testing before re-requesting review for this particular change.

> agent device skills are installed when you install agent-device: https://github.com/callstackincubator/agent-device
> 
> (side note: did we test the full process including the agent-device install on windows?)

I checked transcripts and couldn't see agent-device skill. Digging further it was pretty clear that because we exec through npx, those are not loaded and we can't rely on it. (which is a nice isolation imo, let me know your thoughts)

Windows is not directly tested; assuming devs on Windows are using WSL (You guys probably have a better idea if this is a bad assumption). Will look to validate that in the coming days 

> also, we should probably commit to a single library: `argent` or `agent-device`, instead of recommending both

I'm a little hesitant on this one, imo, going fully blind on it means the agent has no reference when a user brings it up, and that tends to produce more loops and token spend before landing on something maybe works. Let me know if you guys feel strong about this.

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: > should we mention other agents like claude code / codex here?
indeed a good call! Updating

> 
> ideally we should write a script to update the `mcp.json` ARGENT_TOOLS_URL / ARGENT_AUTH_TOKEN for people (could put the script in the skill). the script could detect different agents and also people install mcp under user root or project root

Love the idea! That said both of my sample argent sessions it was able to figure out quickly and configure the MCP on its own without extra instructions. If you are not opposed, would prefer to wait and see if this is a real friction point before adding the script. What do you think?

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: 💡 `simulator:install` incoming, very good idea!

gwdp:
> i have a mixed feeling for the current work. but i don't want to block much. if you want to ship it first and iterate later. please just re-request review for me. (i use graphite dashboard. please don't feel hesitant about the "request changes" / "re-request review" workflow)
> 
> the major concern is
> 
> * some step-by-step in markdowns could potentially use scripts in the skill to make it more deterministic. for example, the argent mcp setup.

I hesitate to give scripts and hard instructions because in my experience, models and user environments are different, and giving the minimum, to allow it to adapt is the part that makes the skill succeed in the most diversity. Happy to be wrong and discuss it further

> * i feel we are mixing "how-to build" concept in this skill. ideally we should keep this skill smaller and link to other skill. currently the differece between mode a/b/c are mostly how the app built. we could try to focus single stuff - "given an app/ipa/apk, use eas simulator" to run it. how to generate app/ipa/apk is not in the skill scope.

Fair, and I agree we opinionated more than ideal. The modes emerged from real dogfooding scenarios though; each one came from a "what if I want to do it this way" moment. I'd rather keep the minimum needed to not leave the model stranded than link out and risk it not finding what it needs in time.

> * imo, mode b is the happy path for me. even though the skill said it's rarely used. if people can build ios app locally, they should mostly have local simulators available. using eas build or eas workflow to build the app, for people running on linux/windows machine to test ios app, is something we are targeting first. maybe we can simplify the mode matrix a little bit.

The framing was mostly to avoid false-triggering on Mac users who just want a local sim. But the trend I'm seeing, even on Mac is the full agent loop: prompt, see, agent sees too. That all-in-IDE experience is where Mode C earns its place as default. For Linux/Windows targeting iOS, yeah Mode B is the right happy path there, agreed 💯 

Thanks for the review once again @Kudo, replies above

@davidmokos dropped some more improvements, if you dont mind a second check, I do appreciate the time 🙇‍♂️

plugins/expo/skills/eas-simulator/SKILL.md
gwdp: Following up 👍

Kudo (approved):
thanks for answering the questions. you know what you want. the current work is based on your thinking. that's great and i don't want to block it. approving the pr to move forward.

leaving answers that we can follow up later 

> Windows using WSL

i think that's not correct. a lot windows user still use powershell / cmd without WSL.

> The framing was mostly to avoid false-triggering on Mac users who just want a local sim. But the trend I'm seeing, even on Mac is the full agent loop: prompt, see, agent sees too. That all-in-IDE experience is where Mode C earns its place as default.

for mac users, if they want agent loop. what's the benefit to use `eas simulator` than local agent-device/argent?

also what does the all-in-IDE experience mean here? does `eas simulator` offer ui panel inside cursor so we can see simulator ui in the same IDE?

gwdp:
Thank you for the quick review @Kudo ! 

> i think that's not correct. a lot windows user still use powershell / cmd without WSL.

appreciate this input! dogfooding some there without WSL and will follow up on this, very interested in making this particular one flawless 

> 
> for mac users, if they want agent loop. what's the benefit to use `eas simulator` than local agent-device/argent?

I see motivation around hardware, but more interesting, reproducing bugs on lower/unsupported iOS versions without having the simulator image installed locally 🥳 (I struggled with this multiple time in the past)

> 
> also what does the all-in-IDE experience mean here? does `eas simulator` offer ui panel inside cursor so we can see simulator ui in the same IDE?

cursor and others are currently offering a built-in browser which is leveraged by the skill -> opens from serve-sim provided tunnel ([check video](https://exponent-internal.slack.com/archives/C0AQJF8BBMZ/p1782369944649689?thread_ts=1782231908.117679&cid=C0AQJF8BBMZ))

Kudo:
thanks for sharing the insight. that makes sense to me.

> I see motivation around hardware, but more interesting, reproducing bugs on lower/unsupported iOS versions without having the simulator image installed locally 🥳 (I struggled with this multiple time in the past)

maybe we can even add this "when to use" to the skill. that could be one selling point.

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,67 @@
...
+
+Alternatively, on a non-sandboxed machine you can use `argent link` instead of env vars:
+
+```bash
+# --no-verify skips the health check if the session isn't fully live yet
+argent link '<ARGENT_TOOLS_URL>' --token '<ARGENT_AUTH_TOKEN>' --yes
+```
+
+**Known issues:**
+- Screenshots crash on EAS worker VMs (`NSPasteboard` error — worker runs under a System LaunchDaemon with no window server). Fall back to agent-device for screenshots.
```
szdziedzic: That's fixed by argent team

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,67 @@
...
+      "command": "argent",
+      "args": ["mcp"],
+      "env": {
+        "ARGENT_TOOLS_URL": "<ARGENT_TOOLS_URL from simulator:start>",
+        "ARGENT_AUTH_TOKEN": "<ARGENT_AUTH_TOKEN from simulator:start>"
+      }
+    }
+  }
+}
+```
```
szdziedzic: I believe we recommend using `argent link` when running `eas simulator:start --type argent`

I think it's easier to manage than editing mcp config file for every tool separately and then remembering to clear these overrides once session is over to make local argent work again?

plugins/expo/skills/eas-simulator/references/controllers.md
szdziedzic: I think this is tricky. If we say to user/agent hey you can use argent/agent-device to interact with the app it's strange that we say "oh but here is this non-standard magical command for EAS that you wouldn't use locally with argent/agent-device that you need to use".

I would rather add support for remote installs to argent and fix the problem properly for everyone.

What do you think?

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: Appreciate highlighting this! Confirmed, dogfooded argent run screenshot against a live EAS session on `0.13.0` and it works cleanly 🙏  Removed the known issue from the skill

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: Flipped it — argent link is now primary 🙏 

I originally went with env vars because they work universally including sandboxed shells (Claude Code) where `~/.argent/` isn't writable, but that's a good catch, it's the minority of cases. Documented env vars as an explicit fallback for those environments. Dogfooded both works well.

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: > oh but here is this non-standard 

I didn't have this perspective, very helpful and I do agree, from the interaction side of things, going for eas-cli `simulator:*` verbs would be non-standard 🤔 Looking to get a feat. on argent and iterate over this skill once that lands. @szdziedzic Let me know if that sounds like a good plan for you

szdziedzic: awesome

plugins/expo/skills/eas-simulator/references/controllers.md
```diff
@@ -0,0 +1,67 @@
...
+- **`press`, not `tap`.** The tap verb is `press` — `tap` is not a verb.
+- **`snapshot -i` is slow on iOS** — tens of seconds is normal; wait for it.
+- **`install` uploads** a local binary to the daemon; **`install-from-source`** has the VM download from a URL (use for EAS artifacts — avoids a large upload).
+- **Proven in this skill's flows:** `apps`, `install`, `install-from-source`, `open`, `snapshot -i`, `press`, `fill`, `screenshot`. Others (`gesture`/`scroll`/`logs`/`record`/`network`/`perf`/`metro`) are real in the CLI but not exercised here — confirm via `<verb> --help` before relying on them.
+
+## argent (alternative)
+
+`npx --yes eas-cli@latest simulator:start --type argent` provisions an argent remote session. The connection config it returns is different (`ARGENT_TOOLS_URL` / `ARGENT_AUTH_TOKEN`).
+
+**argent is a driver, not an installer.** Its `reinstall-app` tool runs `xcrun simctl install` on the remote VM with a literal local path — it does not upload a binary or accept a URL. **Always use agent-device (or `install-from-source`) for the install step**, then use argent to drive the already-installed app.
```
szdziedzic: BTW `*Always use agent-device (or `install-from-source`) for the install step**` won't work for `--type argent` as we don't start agent-device daemon in our VM for this setup. So I think it's safer to just don't describe argent here for now? Or just say you can't install apps in argent mode as of today? I don't want it to be misleading. We can just add argent section here if we add reinstall app suport.

Does it make sense? What is your opinion? 😄

plugins/expo/skills/eas-simulator/references/run-your-app.md
```diff
@@ -0,0 +1,184 @@
...
+```bash
+# 1. Add a simulator build profile to eas.json (merges — won't overwrite existing profiles)
+node -e "
+const fs = require('fs');
+const p = 'eas.json';
+const c = fs.existsSync(p) ? JSON.parse(fs.readFileSync(p, 'utf8')) : {};
+c.build = c.build || {};
+c.build.sim = { ios: { simulator: true } };
+fs.writeFileSync(p, JSON.stringify(c, null, 2) + '\n');
+"
```
szdziedzic: I think it won't work for some setups. It would be good if agent could prioritize getting existing sim profile first before creating new one?

szdziedzic (approved):
Good job 👏 . I think this looks good overall so I'm approving it although I left some comments about argent. Let's release it so we can all test it more easily and see if we like it + iterate on the feedback. Would be great to test it in Cursor Cloud Agents too. I will play with it more once it lands. Awesome.

plugins/expo/skills/eas-simulator/references/controllers.md
gwdp: Good catch 🙇‍♂️ 100% and maybe we can get this limitation sorted out soon. I actually went back through my transcripts/notes thinking I'd tested this workaround, but turns out I was only using EAS builds. 

Updated to clairfy that `argent`sessions have no `agent-device` daemon, so install isn't supported today. Will add it back once there's a real path

plugins/expo/skills/eas-simulator/references/run-your-app.md
gwdp: > I think it won't work for some setups

You're right, there are multiple setup variances, and from experience, the script could force the LLM to use it without much consideration. (reason why I'm not a big fan of script as instructions)

Updated to pure instructions without a script, allowing us to support a little more than the previous iteration.

PR #91, opened 2026-06-24
Add expo-project-structure skill for greenfield app layout

zvadaadam:
## Summary
Adds an `expo-project-structure` skill documenting Expo's recommended folder layout for **brand-new (greenfield)** projects, and lists it in the plugin README.

## The skill
- Covers the SDK 55+ `src/` layout, `src/app` as **routes-only**, and feature-based **colocation** — promote code to shared dirs only once a second feature needs it.
- **Greenfield guidance only.** The leading guardrail tells the agent to follow an existing app's conventions and never restructure to match.
- The SDK version is stated as a durable "since SDK 55" threshold (verified still the default in SDK 56), not a stale "latest version" claim.

## Changes
| File | Change |
|------|--------|
| `plugins/expo/skills/expo-project-structure/SKILL.md` | New skill (+58) |
| `plugins/expo/README.md` | Capability bullet + Skills Included entry |

Validated with `claude plugin validate ./plugins/expo` (passes).


plugins/expo/skills/expo-project-structure/SKILL.md
```diff
@@ -0,0 +1,58 @@
+---
+name: expo-project-structure
+description: Folder structure for a new Expo app — the SDK 55+ `src/` layout, Expo Router `app/` for routes, plus `components/`, `features/`, `hooks/`, `lib/`, `constants/`. Use when scaffolding or laying out a greenfield Expo project, or deciding where a file should live in one. Greenfield guidance only — never restructure an existing app to match.
```
kadikraman: The SDK 55 note is redundant, I believe this is when we changed the default template to use a `/src` folder, but it is supported on router since forever.

```suggestion
description: Folder structure for a new Expo app. Use when scaffolding or laying out a new Expo project using Expo Router, or deciding where a file should live in one. Greenfield guidance only — never restructure an existing app to match.
```

plugins/expo/skills/expo-project-structure/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+version: 1.0.0
+license: MIT
+---
+
+# Expo Project Structure
+
+A starting skeleton for a **greenfield** Expo app — one with no committed folder structure yet.
```
kadikraman: I don't think we should use the word "greenfield" here, I'd stick with "new" throughout (replacing 6 instances in this diff). This is because in our docs we use "greenfield" to mean "an app that was created using Expo or React Native from the start" vs "brownfield" meaning an app that was created using Swift/Kotlin and React Native was added on top (see https://docs.expo.dev/brownfield/overview/ )

```suggestion
A starting skeleton for a **new** Expo app — one with no committed folder structure yet.
```

plugins/expo/skills/expo-project-structure/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+  types/        # Shared TypeScript types
+```
+
+Create a directory on first need — an empty `store/` or `types/` earns its place only when something fills it.
+
+## `src/app` is routes-only
+
+`src/app` holds nothing but route files — screens, `_layout` files, and navigation. Everything else (components, hooks, helpers) lives in the sibling directories above. If both `app/` and `src/app/` exist, Expo Router uses `src/app/` only.
+
+## Colocate by feature; promote only what's shared
```
kadikraman: I've never been a fan of the feature-based colocation personally, but if you feel strongly about it or it's generally better for LLMs, it's okay.

kadikraman (commented):
There's quite a lot missing compared to my [blog post](https://expo.dev/blog/expo-app-folder-structure-best-practices#separate-server-code). This looks more like a recommended folder structure for a web project.

Notable things I definitely suggest adding:

1. a `/screens` folder - in`/app` you can only place files that are screens so components that belong to the screen need to either be in the same file or float in a random `/components` folder otherwise
2. no mention of API routes: call out how to separate [server code](https://expo.dev/blog/expo-app-folder-structure-best-practices#separate-server-code) - only relevant if you're using API routes, but a huge source of confusion
3. where to put your styles
4. where to put your tests
5. where to place platform specific code (as in if you have a different component for ios and android, AI tends to do an inline if/else for it abut using platform specific components is a lot cleaner for bigger changes)

Not in my blog, but something I suggest you add given your expertise: where to put your  AI instruction code like project-specific skills, agents.md / claude.md etc.

plugins/expo/skills/expo-project-structure/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+
+`src/app` is **routes-only**: every file there becomes a route, so nothing else belongs in it. Everything below lives in sibling folders.
+
+## components/ — reusable UI
+
+Generic, reused UI (button, card, table) with one named export each. Name files in **kebab-case** (`bar-chart.tsx`), matching the default `create-expo-app` template. When a component grows, give it its own folder with the root in `index.tsx` and **colocate** its private sub-components beside it — the import path (`@/components/table`) stays unchanged.
+
+## screens/ — screen bodies
+
+Because `app/` files must be routes, complex screen UI that isn't reused has no home there. Put it in `screens/` and let each route just render its screen:
```
kadikraman:
```suggestion
Because `app/` files must be routes, complex screen UI that isn't reused has no home there. Once a screen grows big enough to need breaking out to separate components, put it in `screens/` and let each route just render its screen:
```

plugins/expo/skills/expo-project-structure/SKILL.md
```diff
@@ -0,0 +1,106 @@
...
+  // route-specific concerns only — e.g. read url params here
+  return <Home />;
+}
+```
+
+**Colocate** a screen's private components inside its folder (`screens/home/components/`). A bonus: the same screen can render under multiple routes.
+
+## server/ + app/api/ — separate server code
+
+Appending `+api` to a file in `app/` makes it a server **API route**. Server code is different from frontend code — it runs in a Node-like EAS Hosting environment and can read secret env vars (`process.env.X`, not just `EXPO_PUBLIC_*`). Keep it apart:
```
kadikraman: You can also host API routes on 3rd party services https://docs.expo.dev/router/web/api-routes/#hosting-on-third-party-services

kadikraman (approved):

PR #97, opened 2026-06-30
Improve expo-dev-client skill

amandeepmittal:
## Why

Fix ENG-22278

## How

- expo-dev-client/SKILL.md: lead with dev clients as the recommended setup, keep the concrete "when you need one" list (adding config plugins, remote push, App/Universal Links), and reframe "try Expo Go first" as a quick-test option. Links the docs page as source of truth.
- Bumped Claude/Codex/Cursor plugin manifests 1.3.1 -> 1.3.2 for the version-bump check.

### DIFF
--- a/plugins/expo/skills/expo-dev-client/SKILL.md
+++ b/plugins/expo/skills/expo-dev-client/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: expo-dev-client
-description: Build and distribute Expo development clients locally or via TestFlight
+description: Build Expo app for development
 version: 1.1.0
 license: MIT
 ---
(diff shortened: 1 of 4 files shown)


plugins/expo/skills/expo-dev-client/SKILL.md
```diff
@@ -1,22 +1,22 @@
 ---
 name: expo-dev-client
 description: Build and distribute Expo development clients locally or via TestFlight
```
Kudo:
```suggestion
description: Build Expo app for development
```

nit: `description` does matter for skills. here i see "development clients" is a keyword people would unlikely type in their prompts.

but at the same time, i'm also certain my suggestion is better. maybe to leave it as the original one and follow up later from us.

Kudo (approved):

plugins/expo/skills/expo-dev-client/SKILL.md
amandeepmittal: Let's apply it.

PR #98, opened 2026-07-02
Organize skills with expo-* and eas-*

jonsamp:
# Why

We're adding many skills and our naming conventions and patterns have started to sprawl.

This PR makes some changes to organize them:
- Prepend every skill with either `expo-*` or `eas-*`. This helps discovery since you can type `/expo-` and then you'll see all related skills. Also creates a clear boundary btw free/paid.
- Renames skills to match product names that are more established.
  - building-native-ui → expo-router
  - native-data-fetching → expo-data-fetching
  - use-dom → expo-dom
  - web-to-native → expo-web-to-native
  - add-app-clip → expo-app-clip
  - upgrading-expo → expo-upgrade
  - expo-observe → eas-observe
  - expo-cicd-workflows → eas-workflows
  - expo-deployment → eas-app-stores
    - Previously included web deploy info, separated that out into eas-hosting.
  - expo-api-routes → eas-hosting
 - All skill descriptions have `Framework (OSS)` or `EAS service (paid)`.
 - Adds a contributing guide with patterns documented, like how to test your skill.
 - CI enforces conventions, like skill character limits.

Fixes ENG-23102

# Breaking changes

There are no aliases for the old names for skills cli installations. So anyone who has installed `/building-native-ui` will no longer get updates. I don't think there's a way around this.

# Test plan

Please look at the skils and make sure they seem reasonable. Let me know if there are naming conventions/etc that you think we should change.

The new README is probably the best place to start: https://github.com/expo/skills/blob/bd295c4c5038247d6991aeac1e2d438118df62d1/README.md


Kudo:
i feel building-native-ui is too large and not making sense map to expo-router directly. for instance, the `webgpu-three.md` isn't related to router.

maybe wait #99 to land first so we would be have clear expo-router skill  then

plugins/expo/skills/expo-data-fetching/SKILL.md
```diff
@@ -386,31 +354,7 @@ export {};
...
-
-**With React Query** (automatic):
-
-```tsx
-// React Query automatically cancels requests when queries are invalidated
-// or components unmount
-```
+AbortController on unmount (React Query cancels automatically): see [./references/offline-and-cancellation.md](./references/offline-and-cancellation.md).
 
 ---
```
davidmokos, line 360: <p class='graphite__hidden'><i>[Re: line +444]</i></p>
```gt_suggestion
-> Use EXPO_PUBLIC_ env vars with .env.development and .env.production files
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/expo/skills/98?utm_source=unchanged-line-comment">Graphite</a>.</p>

davidmokos (approved):
I think the new structure is really good. I think some skills are still a bit too long but we should cover that in follow-up PRs

plugins/expo/skills/expo-data-fetching/SKILL.md
jonsamp: Good catch, fixed in 244153e.

Kudo:
late review but i do think we may need some further rename. e.g. expo-native-ui vs expo-ui feels unclear.

PR #101, opened 2026-07-04
Align expo-tailwind-setup with official Nativewind v5 install docs

danstepanov:
This updates expo-tailwind-setup to match the official Nativewind v5 installation docs and fixes a couple of things that have been generating issue reports.

The skill pinned react-native-css to a specific nightly build. Nightlies go stale, fall behind the React Native peer dependency range, and fail install with ERESOLVE on newer Expo SDKs. The install now uses react-native-css@latest and the nativewind preview dist-tag so it does not go stale again, and there is a troubleshooting entry for the ERESOLVE failure.

The setup steps now follow the official docs: peer deps react-native-reanimated and react-native-safe-area-context are included, Tailwind packages install as dev deps, global.css imports nativewind/theme, the CSS file import step is included with the Fast Refresh warning, lightningcss pinning is shown for each package manager, and the TypeScript section covers the nativewind-env.d.ts triple-slash reference. The metro config is plain withNativewind(config) since the default globalClassNamePolyfill gives className support on all React Native components. The useCssElement wrapper setup that was previously presented as required is still there, reframed as an optional section for projects that disable the polyfill.

It also fixes the brand casing across the repo. The project is stylized Nativewind, with a lowercase w.

I maintain Nativewind, happy to answer any questions.


douglowder:
@danstepanov I fixed the conflicts with current main. You need to do two things to fix CI:
- Bump the plugin versions (see `CONTRIBUTING.md`)
- Reduce the size of the tailwind `SKILL.md` to 500 lines or less. Alternatively, you can break it up into references, see for example how `eas-observe` is structured

douglowder (changes requested):
See my previous comment:
- Bump the plugin versions per contributor's guide
- SKILL.md needs to be under 500 lines, or refactored into references

douglowder:
@danstepanov I have created a new branch, `doug/pr101-rebase`, where I have cherry-picked your changes on top of current main, added the required plugin versions bump, and refactored the skill into two references (setup and usage).

Kudo:
actually i was thinking we should not maintain the skill in the expo plugin. rather than that, maybe you should maintain your own skills.

Kudo:
closed by #151. thanks for the contribution anyway. hopefully you could find a more maintainable way to host the skills

douglowder:
@danstepanov following up, after internal discussion, we would like to convert the current Expo skill to a reference that points to your skills, so that you own the source of truth and can make sure it stays up to date with your changes.

PR #105, opened 2026-07-14
Skill eval CI: run eval-experiments harness against every PR touching plugins/expo/skills

kartik-georgian:
## Summary

Skill-eval CI for PRs that touch `plugins/expo/skills/**`.

Runs the [eval-experiments](https://github.com/expo/eval-experiments) harness against all 3 reference PRDs (notes, hot_chocolate, wiki_reader). Each is authored with Claude Code loading the skill plugin **straight from this PR's own checked-out `plugins/expo/` directory** (via a new `SKILL_PLUGIN_DIR` env var → `--plugin-dir`, added in eval-experiments#10), not the published marketplace version — so this actually tests the PR's proposed changes, not what's already released.

## Resolved: pull_request checkout uses the PR's head commit

Confirmed empirically (not documented in Expo's docs or the public eas-cli source, so we tested it directly): opened a real test PR against eval-experiments with a GitHub App connection in place, edited a tracked marker file, and confirmed the automatically-triggered run's checkout reflected the edited value — not the base branch. This confirms `SKILL_PLUGIN_DIR` will work as designed once this repo's connection is live: a real PR's proposed skill changes will actually get checked out and tested, not what's already on `main`.

## Simplified: Expo MCP auth no longer needs a provisioning job at all

mcp.expo.dev now accepts `EXPO_TOKEN` directly as its bearer token (verified live, including a real Claude Code session successfully calling an MCP tool). This removed the whole `provision_mcp_token` job and its `EXPO_MCP_BEARER_TOKEN`/`EXPO_MCP_AUTH_STATUS` wiring from all 3 `author_app_*` jobs — see eval-experiments#15. Submodule pin bumped to track this. `EXPO_MCP_CLIENT_ID`/`EXPO_MCP_REFRESH_TOKEN` secrets on this project are no longer needed (can be cleaned up, not required for function).


.eas/workflows/skill-eval-ci.yml
```diff
@@ -0,0 +1,348 @@
...
+      - uses: eas/checkout
+
+      - name: fetch eval-harness
+        run: |
+          git config --global url."https://x-access-token:${EVAL_HARNESS_ACCESS_TOKEN}@github.com/".insteadOf "https://github.com/"
+          git submodule update --init eval-harness
+
+      - name: author app (Claude Code, this PR's own skill checkout)
+        run: |
+          export SKILL_PLUGIN_DIR="$(pwd)/plugins/expo"
```
kartik-georgian, line 59: We ideally want this to pick up the modified skill files on the developer's PR.

kartik-georgian: This was verified!

Kudo (approved):
looks good for the changes. there's more we can improve as you mentioned at slack, but we can improve that after we have this evals pipeline.

random thought:
- claude code will fetch the plugin from git. hopefully it won't sync the submodule because it's a private repo.
- i'm not sure if the authored app is created through `create-expo-app`. if yes, the app would have `.claude/settings.json` to enable `expo@claude-plugins-official`. i had a setup to disable the plugin: https://github.com/expo/skills/blob/9ce158ff6847c7aaddd3c4cf2dec7eec8a312d3d/.claude/eval-shared/scripts/make-fixture.sh#L92-L102. that would prevent claude mixes installed plugin and `--plugin-dir` plugin

PR #109, opened 2026-07-16
Add expo-migrate-module skill

tsapeta:
# Why

Adds a skill that migrates the Swift side of an existing Expo module from the Expo Modules API 1.0 definition DSL to the 2.0 macro API (`@ExpoModule`, `@JS`, `@Event`, `@SharedObject`, `@Record`) while treating the current JS/TypeScript surface as the compatibility contract.

**Stacked on #113.** Per the review discussion, the 2.0 API is not finalized, so the skill ships from the new `expo-experiments` plugin and graduates to `expo` when the API stabilizes.

# What

The skill, at `plugins/expo-experiments/skills/expo-migrate-module/`:

- `SKILL.md`: four-step workflow (inventory the JS contract, verify the available 2.0 surface in the checked-out source, migrate one semantic group at a time, verify behavior), a `expo@57.0.7` prerequisite check, and a blocked-item flow (ask co-exist vs. revert, open a tracking issue for missing 2.0 coverage).
- `references/migration-map.md`: 1.0-to-2.0 mappings with semantic traps: the three Promise-returning shapes (`async throws`, checked continuation, `Task`-returning), the two paths for queue-pinned functions, `Constant` to `@JS let`, events with explicit wire names, records, shared objects including static properties, and the module lifecycle hook mappings.
- `references/compatibility.md`: how to verify macro declarations and paired core hooks instead of trusting an SDK number, with per-capability evidence gates.
- `references/example.md`: one module walked from 1.0 DSL through mixed mode to fully migrated.
- `agents/openai.yaml`: Codex interface metadata.

Distribution and catalogs:

- `skills.sh.json` gains an **Experimental** group (that channel scans the whole repo and has no exclude mechanism, so experimental skills are listed under an explicit label; the `npx skills` CLI is per-skill opt-in anyway). `check-skill-limits.ts` learns the new group.
- The root README gains a matching Experimental section; the `expo` plugin catalogs are untouched by the skill.
- `expo-module` keeps a small redirect ("Not for migrating... use expo-migrate-module") so migration prompts route correctly for users who have both plugins.
- Versions: `expo` 1.8.3 (its only change is the redirect note), `expo-experiments` stays 0.1.0 (the scaffold version was never released).

# Test Plan

- `claude plugin validate .`, `./plugins/expo`, and `./plugins/expo-experiments` pass.
- `bun scripts/check-skill-limits.ts` passes, including the new Experimental group handling.
- `bun scripts/check-plugin-version-bump.ts origin/main` passes (all three `expo` manifests bumped together).


Kudo (changes requested):
do we need this to be a skill? can it be a reference inside the `expo-module` skill?

tsapeta:
> do we need this to be a skill?

Yes, I think a skill is the fastest way to try the new API in real modules. We'll need to document it eventually, but the API isn't even in beta yet, so I'd hold off on adding it to the public documentation until the surface is more stable.

> can it be a reference inside the `expo-module` skill?

It could. Initially, though, I'd prefer a more targeted and narrowly scoped workflow while the API is still experimental. A separate skill fits that better and gives us room to iterate. We can move it into `expo-module` later once the API stabilizes.

And here is a response from Sol:

> I considered that, but I think this works better as a separate skill. Migrating from 1.0 to 2.0 is an end-to-end, behavior-preserving workflow rather than additional API reference material: it inventories the existing JS contract, checks which v2 capabilities are actually available, handles mixed-mode migration, and verifies parity afterward.
>
> Keeping it separate gives migration requests a precise trigger and keeps expo-module focused on general module authoring. The main SKILL.md is only about 80 lines; detailed mappings and compatibility checks are progressively loaded from references.

Kudo:
i was more asking if it can be references inside expo-module skill. i do see the expo plugin as a whole thing. too many skills would actually loose the coherence. if SKILL.md length is a concern, we can add just one line and link agent to read the references in details.

tsapeta:
I thought about that too, but I think it works better as a separate skill. The trigger is different: "write a module" vs "migrate my module to 2.0", and per `CONTRIBUTING.md` one skill should do one job. If we merge them, the `expo-module` description would have to cover both, and it already triggers well for authoring.

The other thing is that this isn't really reference material. It's a workflow: inventory the JS API first, check what the installed core actually supports, migrate group by group, and there's a defined path for things that can't be migrated yet. A reference file inside `expo-module` would only be read if the agent already knows it's there, and for a migration request nothing in the authoring skill points it that way.

I did add a redirect in [d814308](https://github.com/expo/skills/pull/109/commits/d814308c43c482dc1cc3099420682d9091fe055d) though, so expo-module now explicitly says migration is not its job and points to expo-migrate-module. That should remove the overlap concern.

Kudo:
> per CONTRIBUTING.md one skill should do one job.

that could be over-interpreted. otherwise we should have expo-ui-swift-ui, expo-ui-jetpack-compose, expo-ui-universal, expo-ui-drop-in-replacement for expo-ui 😅

> I did add a redirect in [d814308]
(https://github.com/expo/skills/pull/109/commits/d814308c43c482dc1cc3099420682d9091fe055d) though, so expo-module now explicitly says migration is not its job and points to expo-migrate-module. That should remove the overlap concern.

that's correct to have it on the skill description. the expo-module description could add sentences that the skills also support 2.0 migration. we could link the migration steps in its references

tsapeta:
> that could be over-interpreted. otherwise we should have expo-ui-swift-ui, expo-ui-jetpack-compose, expo-ui-universal, expo-ui-drop-in-replacement for expo-ui

Honestly, I don't think I over-interpreted this. I read "job" as the task to be done - usually you can associate it with a specific verb. In case of `expo-ui`, its job is to **build** UI with `@expo/ui`. SwiftUI/Compose/universal are just flavors of the same job, so that's why they should not be separated.

`expo-migrate-module` job is to **migrate** a module, not to **create** a module. I do think that Expo Modules 2.0 API reference should be part of `expo-module` skill, but we don't have it yet and I don't want to add it during the experimental phase.

I have quite a strong opinion here, but I'm not a maintainer of this repo so the decision is up to you.

Kudo:
> `expo-migrate-module` job is to **migrate** a module, not to **create** a module. I do think that Expo Modules 2.0 API reference should be part of `expo-module` skill, but we don't have it yet and I don't want to add it during the experimental phase.
> 
> 

since expo-module skill doesn't have verb or flavor in its name. create and migrate could be both under the umbrella skill 😅

> 
> I have quite a strong opinion here, but I'm not a maintainer of this repo so the decision is up to you.

alright AI is that nobody can have the correct answer. to move this forward, we can either have evals to prove having expo-migrate-module  doesn't break the plugin coherence, or simpler to add a dedicated "plugins/expo-experiments" plugin, so the change won't affect the plugin under official claude/codex expo plugin.

if it's EM 2.0 isn't finalized, the skill is experimental, i would recommend the latter approach

Kudo (approved):
push to my fork and it works by running `bunx skills use https://github.com/Kudo/expo-skills/tree/expo-migrate-module/plugins/expo-experiments/skills/expo-migrate-module | claude`

plugins/expo-experiments/skills/expo-migrate-module/references/compatibility.md
```diff
@@ -0,0 +1,73 @@
...
+Locate `ExpoModulesMacros.swift` in the target repository or installed dependencies:
+
+```bash
+rg --files | rg 'ExpoModulesMacros\.swift$'
+```
+
+Inspect the declarations that the user's source can import:
+
+```bash
+rg -n 'public macro (ExpoModule|JS|Event|SharedObject|Record|Union|ViewProps|ExpoView)' <path-to-ExpoModulesMacros.swift>
```
Kudo: should not assume rg is available

Kudo (approved):

PR #111, opened 2026-07-17
skill-eval-ci: add main-vs-PR before/after comparison

kartik-georgian:
Adds a main-vs-PR comparison to skill-eval-ci, plus baseline caching

  - Each of the 3 reference PRDs (notes, hot_chocolate, wiki_reader) is now authored and evaluated twice per CI run: once against main's current skill content, once against this PR's — so the comment shows whether a skill change actually improves trigger recall/precision, not just an after-only snapshot.
  - The comment also shows code-quality signals: lexical/structural uptake-check pass rates and syntax/bundle build health.
  - Added baseline caching so the "main" side doesn't get recomputed on every PR: a fingerprint (skill content + harness pin + PRD/scenario + agent/model) gates an exact-match cache restore, and a new trusted workflow (skill-eval-main-baseline.yml) populates that cache on every push to main. PR runs only ever read it.
  - Fixed a symlink-following issue in the fingerprint script (could read arbitrary files or hang on a cyclic symlink).


adi-georgian (approved):
Looks good!

PR #114, opened 2026-07-20
[eas-simulator] check availability first, graceful fallback

gwdp:
## Why

Agents pick up this skill and try EAS Simulator on accounts where it isn't enabled, hitting a confusing mid-flow failure. It should verify access up front and, when unavailable, hand off gracefully instead of dead-ending.

## How

Adds a **Check availability first** step: run `eas simulator:availability` before starting a session (read-only, no billing). If not enabled, don't call `simulator:start` — tell the user and fall back to their local path (local sim / EAS Build). Troubleshooting maps the `not enabled` start error to the same graceful handoff.

Depends on `eas simulator:availability`, which isn't in a released eas-cli yet — the check goes live once that ships.


szdziedzic (commented):
Can we acheive the same without adding new avilability command? Just by understanding eas simulator:startoutput better?

gwdp:
> Can we acheive the same without adding new avilability command? Just by understanding eas simulator:startoutput better?

We do lean on the `simulator:start` output as the fallback here already; the pre-check is on top of that to fail fast — since some modes build on EAS first (~15-20 min), relying only on the start error means the user waits through a build before finding out. The check bails right at the start for a nicer experience. Totally open to dropping it if you feel the extra command isn't worth it but focus was in the customer experience, so folks dont get billed with builds to later find they were pursuing a dead path.

szdziedzic (approved):

gwdp:
Confirmed after cli release, polite refusal + nudge into local direction happen once/if flag is disabled on account;

Transcript (final chunk only):
```Availability came back false for gwdp; checking which Expo account this project uses and whether another account has access.

[toolcall]

EAS Simulator isn’t available on the gwdp account that owns this project ({"available": false}), so I won’t start a session — it would fail and shouldn’t be billed.

You’re on macOS with a local simulator available. I can run the app locally with expo run:ios (or Expo Go) and screenshot the home screen instead — say the word and I’ll do that.```

PR #115, opened 2026-07-20
[eas-workflows] document workflow file size limit

gwdp:
Documents the 16 KiB per-file workflow limit in the eas-workflows skill so generated files stay under it. Follow-up to expo/universe#29253 and the docs change in expo/expo#47997.


Kudo (approved):

plugins/expo/.cursor-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "expo",
   "displayName": "Expo",
-  "version": "1.9.4",
```
sjchmiela, line 4: isn't it weird this is part of a pull request?

sjchmiela (approved):

plugins/expo/.cursor-plugin/plugin.json
gwdp: quite annoying 👍 we will get into it

PR #119, opened 2026-07-23
[eas-simulator] move to expo-experiments plugin

gwdp:
## Why

EAS Simulator is a limited-access, experimental feature (`simulator:*` commands are hidden/experimental, and it's not enabled on most accounts). Shipping it in the default `expo` plugin means agents discover and attempt it on accounts without access — the source of the recent support tickets. Moving it to the opt-in `expo-experiments` plugin stops the accidental discovery at the distribution layer; it graduates back to `expo` at GA.

Stacked on #114 (availability check + graceful fallback), which is the runtime safety net for opted-in users. This PR is the distribution-layer half.

## How

`git mv` the skill from `plugins/expo` to `plugins/expo-experiments` (pure move, no content change), and update the catalogs: `skills.sh.json` (Services → Experimental group), root and plugin READMEs. `expo` 1.8.5 → 1.8.6, `expo-experiments` 0.1.0 → 0.2.0.

## Test plan

`check-plugin-version-bump.ts`, `check-skill-limits.ts`, and `claude plugin validate` on both plugins pass.


gwdp:
After chatting with the team, we're going to address this through [#114](https://github.com/expo/skills/pull/114) (the availability check + graceful handoff) instead — that keeps eas-simulator visible in the mainstream expo plugin rather than tucking it into experiments, while still failing fast + degrading nicely for accounts without access. So this move isn't needed. Closing it out 🙏

PR #120, opened 2026-07-24
[eas-simulator] flag the first-time deep-link Open dialog

gwdp:
## Why

The first time an agent deep-links into the app, iOS pops an "Open in '<app>'?" dialog it has to tap. Without flagging it, the agent snapshots the screen, notices the dialog, then decides to press it — a wasted round trip on the first deep-link of every session.

## How

- agent-device: expect the dialog on a first deep link and press "Open" straight away.
- Fixed a macOS timeout gotcha I hit while dogfooding — the guidance wrapped the press in a shell `timeout`, but macOS has no `timeout` binary, so it'd bail. Uses agent-device's own `--timeout` now.
- Covered the argent path too: its UI queries can miss system dialogs, so follow argent's own hint to make them visible, then tap Open.

## Test Plan

Dogfooded both controllers end to end — agent-device (normal launch, no stray Open press) and argent (deep-link → dialog → tap → landing). Skill checks pass.


szdziedzic (approved):

gwdp:
@szdziedzic dogfooding this more today and hit two things worth folding in since it wasn't merged yet: 
- macOS timeout-handling issue -> the guidance said to wrap the press in a shell timeout, but macOS has no timeout binary, so it'd bail; switched it to agent-device's own --timeout.
- tested the argent path too and documented how it handles the system Open dialog (needs boot-device --force before describe can see it, then gesture-tap). 

Feels like both improvements on timeout/dialog handling, so made sense to amend here. Re-requesting review 🙏

szdziedzic (approved):

PR #125, opened 2026-07-30
[eas-simulator] always pass --name to simulator:start

brentvatne:
Unnamed sessions show as `unnamed` in `eas simulator:list` / `eas simulator:get`, and every row of the **Simulator sessions** page on expo.dev reads "Simulator session" over a random id — a wall of identical entries nobody can navigate.

This teaches the skill to always pass `--name` on `simulator:start`, and to generate a name a human can recognize later — derived from what the user asked for, not from the mode or the tooling.

- **Length: aim 3–6 words / ~40 characters, 50 practical.** The name renders as a single-line title in a narrow table column. The **255-character** hard cap and the rejection of empty/whitespace-only names are verified from the GraphQL schema in `eas-cli@21.4.0` (`CreateDeviceRunSessionInput.name`); the shorter target is a style call.
- **Sentence case.** The name replaces "Simulator session" as the row title, which is itself sentence case. Identifiers keep their real casing (`Dev build for expo-router v4`).
- **No metadata the table already shows.** Each row already displays the session id, platform, start time, duration, and creator, so the whole budget goes to purpose.

### DIFF
--- a/plugins/expo/.claude-plugin/plugin.json
+++ b/plugins/expo/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "expo",
-  "version": "1.8.6",
+  "version": "1.8.7",
   "description": "Official Expo skills for building, deploying, upgrading, and debugging Expo apps",
   "author": {
     "name": "Expo Team",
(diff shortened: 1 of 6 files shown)


gwdp (approved):
Since the parameter is only available in newer CLI versions and could potentially cause `simulator:start` failures, it would be relevant to include instructions in the troubleshooting section. For example, if `--name` is not recognized, suggest updating the eas-cli or omitting it.

PR #126, opened 2026-08-04
[skills] collect eval-candidate signals via new evals feedback category

zvadaadam:
## Summary

Teaches agents to report Expo tasks that AI models fail at as **eval candidates**, riding the existing `submit-expo-feedback` pipe with a new `evals` category. These structured leads are the raw material for Harbor eval tasks — the hard, real-world, model-breaking tasks that are otherwise nearly impossible to source. Anything Expo an agent can attempt qualifies (framework, EAS, tooling), whether or not a skill was involved: if we can eval it, we can improve it.

## Changes

| Area | Change |
| --- | --- |
| `expo-skill-feedback` | New "Eval candidates" section with a structured lead template (Task / Environment / Expected / Actual / Wrong approach / Evidence), `evals` row in the category table (subject = Expo area or capability, e.g. `expo-router`, `eas build`), eval-candidate trigger added to the description, Codex `openai.yaml` updated |
| All skills | Canonical feedback footer gains one line routing agent failures to the eval-candidate flow (template in `check-skill-limits.ts`, propagated via `--fix-feedback`) |
| Manifests | Version bump `1.8.7` → `1.9.0` across Claude, Codex, and Cursor manifests |

## Notes

- The template fields map onto Harbor task anatomy: Task → `instruction.md`, Environment → `environment/`, Expected → rubric criteria, Wrong approach → "Must not …" negative criteria, Evidence → difficulty/solvability calibration.
- No backend change needed: `/v2/feedback/cli-send` passes categories through untouched to PostHog (`cli_feedback_submitted`) and Slack, so `evals` submissions are filterable server-side today.
- `--category evals` requires a `submit-expo-feedback` release (the category enum is client-side; companion change prepared for expo/expo). Until it ships, the skill documents the `--category unknown --subject "eval-candidate: <area>"` fallback.

## Validation

- `bun scripts/check-skill-limits.ts` ✓
- `bun scripts/check-plugin-version-bump.ts origin/main` ✓ (1.8.7 → 1.9.0, all three manifests together)
- `claude plugin validate .` and `claude plugin validate ./plugins/expo` ✓

PR #128, opened 2026-08-04
[skills] drop evals category fallback now that the CLI supports it

zvadaadam:
## Summary

Removes the interim fallback from the eval-candidate flow (`--category unknown --subject "eval-candidate: <area>"`), per @davidmokos's review comment on #126 — the `evals` category is now supported by the CLI itself (expo/expo#48511). A minimal error rule stays in its place ("if the command fails, report the error to the user instead of resending") so agents don't improvise resends on transient errors, which the usage simulation showed they otherwise do.

## Merge gate

**Do not merge until `submit-expo-feedback` ≥ 0.0.3 is published to the npm `latest` dist-tag** (currently `latest` = 0.0.2, which still rejects `--category evals`). The skills invoke `npx --yes submit-expo-feedback@latest`, so the release — not the expo/expo merge — is what makes the primary path work; until then this fallback is agents' only working path for eval submissions.

## Validation

- `bun scripts/check-skill-limits.ts` ✓
- `bun scripts/check-plugin-version-bump.ts origin/main` ✓ (1.9.0 → 1.9.1, all three manifests)
- `claude plugin validate ./plugins/expo` ✓

**Update:** also syncs the category table with the CLI's new contract from expo/expo#48520 (adds the `simulator` row; `evals` wording already matches). Both changes ride the same release gate.

### DIFF
--- a/plugins/expo/.claude-plugin/plugin.json
+++ b/plugins/expo/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "expo",
-  "version": "1.9.0",
+  "version": "1.9.1",
   "description": "Official Expo skills for building, deploying, upgrading, and debugging Expo apps",
   "author": {
     "name": "Expo Team",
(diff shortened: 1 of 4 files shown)


davidmokos (approved):
good, merge when the new submit-expo-feedback cli is released in stable

PR #137, opened 2026-08-06
[eas-simulator] agent-device screenshot --scale, --max-size removed

gwdp:
## Why

agent-device [PR #1617](https://github.com/callstack/agent-device/pull/1617) adds a scale-only `screenshot --scale` knob and removes `--max-size`. Our controller reference told agents the opposite: that agent-device screenshots are "full-resolution with no scale knob."

## How

- Rewrote the agent-device screenshot note in `references/controllers.md`:
  - `screenshot --scale <0.01–1>` resizes both dimensions, `1` = full resolution.
  - Token-conscious default via `AGENT_DEVICE_SCREENSHOT_SCALE` / `screenshotScale` in config, overridden by an explicit `--scale`.
  - `--max-size` is removed and older calls are refused with migration guidance.
  - Added a remote-session caveat: the resize runs on the daemon, so a newer client against an older EAS session daemon can have `--scale` silently ignored.
- Gated the whole thing on `v0.20.6` as a single token, easy to bump if it ships as a different version.
- Bumped the plugin version to 1.9.7.

## Test Plan

- Cross-check every claim against agent-device#1617's own doc changes: flag name, the `0.01–1` range, the env var / config key, and the `--max-size` removal


szdziedzic (approved):
awesome

PR #143, opened 2026-08-11
[expo-design] add Expo design systems skill

SchroederNathan:
## What

Adds `expo-design`, a skill for building design systems in Expo apps:

- Token theme under one entry point (`src/theme`): colors, spacing, typography, radius, shadows, motion
- Component contract: `variant`/`size` props, pressed/disabled/loading states, `style` merged last
- Rules for when a repeated view earns extraction into `src/components/`
- `references/audit.md`: grep checks for design-system drift (hardcoded style values) + component doc templates

Defers to siblings instead of duplicating: colors → `expo-native-ui`, Tailwind → `expo-tailwind-setup`, folders → `expo-project-structure`.

## Housekeeping

Registered in all catalogs, added `agents/openai.yaml`, bumped plugin 1.9.6 → 1.10.0.

## Testing

CI scripts and `claude plugin validate` pass. Eval harness (2 prompts × with/without skill, SDK 57): all runs pass `tsc`/lint/`expo export` and all graded expectations; with the skill, output kept every visual literal inside the theme, while the baseline leaked hardcoded `fontSize` values.


zvadaadam:
great ideas in here, @SchroederNathan!

how about we call this skill `expo-design-system`? imho better describes what the skill actually does.

SchroederNathan:
> great ideas in here, @SchroederNathan!
> 
> how about we call this skill `expo-design-system`? imho better describes what the skill actually does.

I like it! +1

zvadaadam:
just an idea to consider... what should we when the agent lands inside an app that already has its own design system?

maybe we could add this one section in the begging of the skill? wdyt?

> 
> ## Adopt Before You Build
> 
> In an app that already has screens, the first move is detection, not construction:
> 
> Find the existing system: a theme/tokens file (theme.ts, constants/Colors.ts), a styling library (NativeWind/Tailwind → expo-tailwind-setup, Tamagui, Restyle, Unistyles), or just the de facto values repeated across screens.
> If one exists, it is the source of truth — extend it in its own idiom (its names, its scale, its storage format), and audit drift against it, not against the examples below.
> Only when nothing exists do the defaults below apply as written.
> Never introduce a second system beside an existing one — a fresh src/theme/ next to a Tamagui config is design-system drift.

SchroederNathan:
> just an idea to consider... what should we when the agent lands inside an app that already has its own design system?
> 
> maybe we could add this one section in the begging of the skill? wdyt?
> 
> > ## Adopt Before You Build
> > In an app that already has screens, the first move is detection, not construction:
> > Find the existing system: a theme/tokens file (theme.ts, constants/Colors.ts), a styling library (NativeWind/Tailwind → expo-tailwind-setup, Tamagui, Restyle, Unistyles), or just the de facto values repeated across screens.
> > If one exists, it is the source of truth — extend it in its own idiom (its names, its scale, its storage format), and audit drift against it, not against the examples below.
> > Only when nothing exists do the defaults below apply as written.
> > Never introduce a second system beside an existing one — a fresh src/theme/ next to a Tamagui config is design-system drift.

Good point, I've added it with a change. Also added a `package.json` check for the styling library.
https://github.com/expo/skills/pull/143/changes/5a15ad0a4c6096607979b50b6d5d543ab69deec4

zvadaadam (approved):

Kudo:
oh no the eval-harness had been reverted. i'll update in #145

PR #144, opened 2026-08-12
fix(eas-app-stores): affirm first-time eas submit works on Google Play

amandeepmittal:
# Why

Fix ENG-25835

# How 

Add a statement to affirm first-time eas submit works on Google Play so that it complies with EAS Submit docs here: https://docs.expo.dev/submit/android/

# Test plan

Proofread


kadikraman (approved):
Fun fact, I asked Claude sth about store releases yesterday, and it spotted this change!

Image

PR #145, opened 2026-08-13
update eval-harness

Kudo:
# Why

try eval-harness rewrite for typescript

# How

- update git submodule and fix the workflow yaml
- use sonnet[1m] so that claude would load skill description correctly

# Test Plan

run evals


adi-georgian (approved):
Looks great Kudo! Thanks for this!

PR #148, opened 2026-08-13
chore: add --set-version and --help to the plugin version script

douglowder:
## Why

The three plugin manifests have to be bumped together, to the same version, above the base ref. Editing them by hand is easy to get wrong, and the script that enforces the rule had no way to write the value or explain itself.

## How

Add `--set-version <version>`, which validates the version and writes it to all three manifests. It refuses a version that is not valid semver or is not greater than the base ref, validates every manifest before writing any of them, and replaces only the version value so file formatting is untouched. Add `--help` describing the rule, the watched paths, the flags, and the summary env var.

Replace the isSemver check, which used semver.satisfies(v, v) and so accepted v1.2.3 and 1.2.3.4, with the official semver regex.

## Test plan

Add `scripts/test-plugin-version-bump.sh` covering all 21 cases. It backs up the manifests and restores them on exit, and compares against the backups rather than git, so it passes whether or not the working tree is clean.


douglowder:
* **#148** <a href="https://app.graphite.com/github/pr/expo/skills/148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/expo/skills/148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

Kudo (changes requested):
thanks for helping on this. since we have bun, what do you think to rewrite the check-plugin-version-bump test using bun directly?
https://bun.com/docs/test

davidmokos (approved):
This is a good idea. Agree with the @Kudo's comment, having a bun script would be cleaner

douglowder:
@Kudo @davidmokos test script converted to a bun script, and added a step in the `plugin-version-bump.yml` workflow to run the test on PRs.

scripts/test-plugin-version-bump.ts
```diff
@@ -0,0 +1,210 @@
...
+
+function check(name: string, condition: boolean, detail: string) {
+  if (condition) {
+    pass(name);
+  } else {
+    fail(name, detail);
+  }
+}
+
+function expect(name: string, expectedCode: number, expectedText: string, args: string[]) {
```
Kudo: could we reuse test infra from bun rather than the handy pass/fail/check/expect? it also offers describe so you can split different test suites

https://bun.com/docs/test

```js
import { expect, test } from "bun:test";

test("2 + 2", () => {
  expect(2 + 2).toBe(4);
});
```

Kudo (changes requested):

scripts/test-plugin-version-bump.ts
douglowder: Done

Kudo (approved):

PR #151, opened 2026-08-14
[expo] remove NativeWind skill

davidmokos:
## Why

NativeWind-specific agent guidance should be owned and maintained by the NativeWind team.

## What changed

- remove the `expo-tailwind-setup` skill and its Codex metadata
- remove the skill from every Expo catalog and README
- bump all Expo plugin manifests from 1.10.1 to 1.10.2


Kudo (approved):
thanks for helping on this

plugins/expo/skills/expo-examples/SKILL.md
```diff
@@ -87,7 +87,6 @@ When the user already has an app, **add only what the example introduces; never
 
 ## Related skills
 
-- Tailwind / NativeWind styling → `expo-tailwind-setup`
```
douglowder, line 90: I don't agree that we should remove the skill entirely. I would prefer to

- Wait for the NativeWind team to publish their skill
- Convert `expo-tailwind-setup` into a small skill that points to the NativeWind skill, and optionally adds any Expo specific configuration needed

douglowder (changes requested):

douglowder (approved):
Since there has been no activity on the NativeWind side, let's move forward with this PR to remove the old skill. Once a new NativeWind v5 skill becomes available, we can create a new PR to reference it. @Kudo

douglowder:
@davidmokos @Kudo I approved this and rebased against main, and bumped the plugin versions.

Kudo:
thanks David and Doug. Let's merge this removal pr and may look back to add nativewind as a reference (rather than skill) later

zvadaadam:
awesome, this solves the reported nativewind problem in the [ENG-25794](https://linear.app/expo/issue/ENG-25794/nativewind-skill-uses-outdated-react-native-css-pin-and-old-api).

I'm closing the [#174](https://github.com/expo/skills/pull/174) PR and let's merge this.

PR #155, opened 2026-08-18
[skills] add expo animation skill and upstream sync

davidmokos:
Emil Kowalski created the `animate-expo` skill (https://github.com/emilkowalski/skills/pull/25) in collaboration with Expo.

I want to copy the skill in our skills so our users can create beautiful animations in expo.

This PR adds the `expo-animation` skill with attribution.
It also adds `scripts/sync-animate-expo.ts` to pull future upstream changes while preserving Expo-specific metadata, attribution, and feedback instructions.

Bumps the Expo plugin manifests to 1.11.0.

### DIFF
--- a/plugins/expo/skills/expo-animation/RECIPES.md
+++ b/plugins/expo/skills/expo-animation/RECIPES.md
@@ -1,0 +1,385 @@
+# Expo Animation Recipes
+
+Ready-to-build implementations for the cases that come up most in a React Native app. Start from the recipe, then adapt.
+
+---
+
+## Setup the recipes assume
+
+```bash
+npx expo install react-native-reanimated react-native-worklets react-native-gesture-handler expo-haptics
+```
+
+(`react-native-keyboard-controller` only for the keyboard recipe.) `expo install`, not `npm install` — it resolves the versions that match the SDK. The worklets Babel plugin is configured by `babel-preset-expo` automatically.
+
+`GestureHandlerRootView` wraps the app once — in Expo Router, the root `_layout`:
+
+```jsx
+import { GestureHandlerRootView } from 'react-native-gesture-handler';
+
+export default function RootLayout() {
+  return (
+    <GestureHandlerRootView style={{ flex: 1 }}>
+      <Stack />
+    </GestureHandlerRootView>
+  );
+}
+```
+
+Imports and constants every recipe below shares:
+
+```js
+import { useState, useEffect, useMemo } from 'react';
+import Animated, {
+  useSharedValue, useAnimatedStyle, useAnimatedScrollHandler, useAnimatedReaction,
+  withSpring, withTiming, interpolate, Extrapolation, Easing,
+  FadeInDown, FadeOutDown, LinearTransition,
+} from 'react-native-reanimated';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
+import { scheduleOnRN } from 'react-native-worklets';
+import * as Haptics from 'expo-haptics';
+
+const EASE_OUT = Easing.bezier(0.23, 1, 0.32, 1);      // strong ease-out for UI
+const EASE_IN_OUT = Easing.bezier(0.77, 0, 0.175, 1);  // on-screen movement
+const EASE_SHEET = Easing.bezier(0.32, 0.72, 0, 1);    // iOS sheet curve
+```
+
+Three conventions, explained once here instead of in every recipe:
+
+- **Shared values are read and written with `.get()` / `.set()`**, the form the Reanimated docs recommend for React Compiler support. `.value` still works, but the compiler can't see through it.
+- **`scheduleOnRN(fn, ...args)` replaces the deprecated `runOnJS(fn)(...args)`** for calling back to the React Native runtime from a worklet.
+- **Gestures are wrapped in `useMemo`.** Rebuilding a gesture on every render can reattach the recognizer and drop a drag that's mid-flight.
+
+**Gesture Handler v3:** Expo installs v2, and the recipes use its `Gesture.Pan()` builder. If the project is already on v3, the builder is legacy — each gesture is a hook taking one config object, with `onStart` → `onActivate`, `onEnd` → `onDeactivate`, and the `success` flag replaced by `event.canceled` (inverted). The hook manages its own identity, so drop the `useMemo`:
+
+```jsx
+const pan = usePanGesture({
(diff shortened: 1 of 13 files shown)


plugins/expo/skills/expo-animation/RECIPES.md
```diff
@@ -0,0 +1,385 @@
...
+
+---
+
+## Setup the recipes assume
+
+```bash
+npx expo install react-native-reanimated react-native-worklets react-native-gesture-handler expo-haptics
+```
+
+(`react-native-keyboard-controller` only for the keyboard recipe.) `expo install`, not `npm install` — it resolves the versions that match the SDK. The worklets Babel plugin is configured by `babel-preset-expo` automatically.
```
Kudo, line 13:
```suggestion
(`react-native-keyboard-controller` only for the keyboard recipe.) `expo install`, not `npm install` — it resolves the versions that match the SDK.
```

feels not necessary to mention that one which is configured by default

plugins/expo/skills/expo-animation/RECIPES.md
```diff
@@ -0,0 +1,385 @@
...
+
+---
+
+## Press feedback
+
+Every pressable in the app. This passes the frequency gate only because it's near-imperceptible: 120ms and a 3% scale is the ceiling for something touched this often — anything longer or larger belongs to rarer moments, per step 1 in SKILL.md. No gesture, no shared value — a CSS transition is the whole implementation.
+
+```jsx
+import Animated from 'react-native-reanimated';
+import { Pressable, StyleSheet } from 'react-native';
```
Kudo, line 93: nit: since we have using react-native-gesture-handler, it also offers a Pressable https://docs.swmansion.com/react-native-gesture-handler/docs/components/pressable/

not necessary to use it in the skill though

plugins/expo/skills/expo-animation/RECIPES.md
```diff
@@ -0,0 +1,385 @@
...
+  pressed: { transform: [{ scale: 0.97 }] },
+});
+```
+
+`setState` is fine here — it fires twice per press, not per frame. `hitSlop` brings a small icon up to the 44pt target without growing it; `pressRetentionOffset` stops a slight finger drift from cancelling.
+
+---
+
+## Bottom sheet you can drag to dismiss
+
```
Kudo, line 126: could we use expo-ui BottomSheet?

plugins/expo/skills/expo-animation/RECIPES.md
Kudo, line 13: if keyboard-controller is opt-in and given `npx expo install react-native-keyboard-controller` is mentioned below. maybe we could not mention it here?

plugins/expo/skills/expo-animation/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+description: Framework (OSS). Build animations in React Native and Expo, making the decisions in the order that determines whether they feel right — should it animate, which thread it runs on, which properties, spring or timing, how the gesture hands off, how it degrades. Writes the implementation with Reanimated, Gesture Handler, Expo Router and expo-haptics. Use when animating anything in an Expo app, adding gestures, sheets, screen transitions, press feedback or haptics, or fixing motion that stutters on device. For web animation use `animate`.
+version: 1.0.0
+license: MIT
+---
+
+# Building Animations in Expo
+
+This skill was created in collaboration with [Emil Kowalski](https://github.com/emilkowalski) and can also be found in the [emilkowalski/skills](https://github.com/emilkowalski/skills) repository, along with other useful animation skills.
+
+A construction skill for React Native. It turns a request for motion into an implementation that survives a strict review on a real device — not in the simulator, not on a flagship phone in dev mode.
```
Kudo, line 12:
```suggestion
A construction skill for React Native and Expo. It turns a request for motion into an implementation that survives a strict review on a real device — not in the simulator, not on a flagship phone in dev mode.
```

since the setup recept assumes expo?

plugins/expo/skills/expo-animation/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+
+// or let each animation decide
+withSpring(0, { duration: 300, dampingRatio: 0.8, reduceMotion: ReduceMotion.System });
+```
+
+Reduced motion means **fewer and gentler**, not zero: keep opacity and color changes that explain a state change, drop translation, scale, parallax and overshoot. Screen transitions become `animation: 'fade'`.
+
+**Text scales.** `allowFontScaling` is on by default, so any height you measured at default type size is wrong at 200%. Never animate to a hardcoded height — measure with `onLayout`, or animate a transform instead.
+
+## Setup that silently breaks motion
```
Kudo, line 202: just link to RECIPES.md since there's setup section?

plugins/expo/skills/expo-animation/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+
+Check these first when "the animation just doesn't run":
+
+- Install through Expo so versions match the SDK: `npx expo install react-native-reanimated react-native-worklets`. In an Expo project, `babel-preset-expo` configures the worklets Babel plugin automatically — no `babel.config.js` step. Only a bare RN project without that preset adds the plugin manually, and there it must be last in the list. A missing or misplaced plugin doesn't silently fall back anymore — it throws `Failed to create a worklet` at runtime.
+- `GestureHandlerRootView` must wrap the app, or gestures do nothing with no error.
+- Reanimated 4 requires the New Architecture.
+- **Expo Go is not a performance environment.** Judge feel in a release build; a dev build's JS thread is slow enough to hide exactly the problems you're looking for.
+
+## 120fps
```
Kudo, line 211: should also move to setup section in RECIPE.md?

Kudo (approved):
left some comments but i know it's synced from upstream. leave it for your call and could follow up later if there's any comment makes sense to you

plugins/expo/skills/expo-animation/RECIPES.md
davidmokos: it used to be necessary to manually add the babel plugin and models tend to do it and mess up. I think it's good to keep that info for the more basic models.

plugins/expo/skills/expo-animation/RECIPES.md
davidmokos: isn't it the same as Stack with `presentation: 'formSheet'` as mentioned here?

PR #159, opened 2026-08-19
[eas-app-stores] fix invalid submit job examples

neutronm:
# Why

The `submit` job examples in the `eas-app-stores` skill use `platform:` in
`params`, but the EAS Workflows schema requires `build_id` for submit jobs and
does not allow `platform`.

Expo's published workflow documentation already uses `build_id` wired from the
preceding build job. This brings the skill references in line with that syntax.

# How

- Updated both submit jobs in `references/workflows.md` to use
  `build_id: ${{ needs.<build_job>.outputs.build_id }}`.
- Updated the submit job in `references/ios-app-store.md` the same way.
- Renamed the Production Release job IDs to underscores so the new
  `needs.<job_id>.outputs.build_id` expressions follow the documented EAS
  Workflows job-ID syntax.
- Bumped the Expo plugin version to `1.11.4`.

# Test Plan

- Validated both corrected workflow examples against the current EAS Workflows
  schema.
- Confirmed the original `platform:` form fails validation while the corrected
  `build_id` form passes.
- Ran `bun scripts/check-skill-limits.ts`.
- Ran `bun scripts/check-plugin-version-bump.ts origin/main`.

### DIFF
--- a/plugins/expo/.claude-plugin/plugin.json
+++ b/plugins/expo/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "expo",
-  "version": "1.11.0",
+  "version": "1.11.4",
   "description": "Official Expo skills for building, deploying, upgrading, and debugging Expo apps",
   "author": {
     "name": "Expo Team",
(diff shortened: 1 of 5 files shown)


neutronm:
Hey @Kudo — would you mind taking a look at this when you get a chance? It fixes a few invalid EAS submit workflow examples in the `eas-app-stores` skill. Thanks!

PR #176, opened 2026-09-01
[expo] Correctness pass on design-surface skills

zvadaadam:
## Why

This repo's rule is one canonical home per topic. For motion, that home is `expo-animation` — and it's good: it teaches the animate-or-not gate, sane durations, interruption handling, reduced motion. But `expo-native-ui` still carried an `animations.md` written before that skill existed, teaching the opposite (menus of Bounce/Zoom entrance presets, 700ms entrances, "add entering/exiting animations for state changes"), and `expo-web-to-native` was actively routing motion work to the old file.

Skills are loaded by trigger, so which advice an agent got was a coin flip on entry point. That's a big part of why generated apps animate like old Dribbble shots instead of like iOS. The fix is not more guidance — it's deleting the duplicate and referencing the canonical skill.

The rest of the PR applies the same idea to smaller cases we found while verifying:

- where a skill contradicted **itself**, make it agree (SKILL.md banned `expo-symbols` while all three of its references use it; hex-color snippets sat directly under an "avoid custom colors" rule),
- where a claim was **verifiably false**, correct it against docs or types (Expo Go does not support all `expo-*` packages — Android remote push is gone since SDK 53; `Stepper` doesn't exist in React Native; the stack docs only document `headerLargeTitleEnabled`, not `headerLargeTitle`; tabs.md violated the skill's own no-direct-`@react-navigation` rule).

Agents copy examples, not doctrine — so the examples have to agree with the doctrine. Net −245 lines.

## Changes

**Motion has one home**
- `expo-native-ui/references/animations.md` → short pointer to `expo-animation`, keeping only the native-UI-specific caveat (semantic colors can't enter Reanimated styles)
- `expo-native-ui/SKILL.md` styling rule and `expo-web-to-native/references/native-patterns.md` now route motion work to `expo-animation`

**Wrong APIs and claims**
- Fake `Stepper` import removed, with a guard line so it doesn't get hallucinated back
- `headerLargeTitle` → `headerLargeTitleEnabled` in expo-router examples
- tabs.md `ThemeProvider` now imported from `expo-router/react-navigation` with the correct `value=` prop
- Expo Go support claim made accurate; "always wrap root in ScrollView" scoped (FlatList is its own scroll container; camera/map screens exempt)

**Internal consistency**
- SF Symbols standardized on `expo-symbols` (what every reference already used), plus an Apple-only note with the Android Material-icon rule
- Control snippets no longer recolor native controls under a rule that says not to; TextInput example uses theme semantic colors (`secondarySystemBackground` added to the palette)
- Camera/audio examples: `TouchableOpacity` → `Pressable`, mic permission requested on first record instead of on mount, dead imports removed

**Liquid Glass contract completed**
- Guard `isGlassEffectAPIAvailable()` in addition to `isLiquidGlassAvailable()` — some iOS 26 betas lack the API and crash without it (expo/expo#40911)
- Reduce Transparency → solid surface fallback; never animate opacity on a `GlassView`; `overflow: 'hidden'` stays a BlurView rule and is now explicitly banned above interactive glass

## Verification

`headerLargeTitleEnabled` checked against the official expo-router stack docs; glass guards checked against the installed `expo-glass-effect` type declarations. `check-skill-limits` ✓, `check-overview-routing` ✓, `claude plugin validate` ✓, version 1.12.4 → 1.12.5 in all four manifests.

Follow-ups (separate PRs): native-slop anti-pattern vocabulary, loading/empty-state doctrine, expo-animation interruption/exit craft, per-app design contract + native-feel evals.


plugins/expo/skills/expo-native-ui/references/animations.md
davidmokos, line 1: should we remove this whole file and reference the expo-animation skill from the parent directly? It seems that now it doesn't bring any value

plugins/expo/skills/expo-router/SKILL.md
```diff
@@ -217,13 +217,13 @@ export default function Layout({ segment }) {
         headerLargeTitleShadowVisible: false,
         headerLargeStyle: { backgroundColor: "transparent" },
         headerTitleStyle: { color: colors.label },
-        headerLargeTitle: true,
+        headerLargeTitleEnabled: true,
```
davidmokos, line 220: what if people use this skill with older SDK version though? Do we always assume the latest SDK?

plugins/expo/skills/expo-native-ui/references/animations.md
zvadaadam: yeah, good point! it's gone. same thinking can be applied to expo-native-ui. 

we now only keep one-liner at the top pointing to expo-animation so the agents know about it.

plugins/expo/skills/expo-router/SKILL.md
zvadaadam: hmmm, I'd rather optimize our skills for the latest SDK version and nudge the agent to use our expo-upgrade to upgrade.

we could mention this in the expo-overview: 

> “Skills target the current Expo SDK. Version-gated APIs are marked SDK NN+; for older projects, check the installed package’s types before using a marked option, or run expo-upgrade.”

wdyt? @davidmokos

Kudo: that brings a broader question. i think our skills are assumed latest sdk in difference places. how could we support sdk versioned skills? not sure if co-located skills would help since we are publishing skills within npm packages

Kudo (approved):

PR #178, opened 2026-09-01
[expo] native-slop tells, four-state screens, Dynamic Type, keyboard, Android floor

zvadaadam:
Closes the top silent zones from the native-feel audit: the places where the skills say nothing and generated apps fall back to web habits. Also names the recurring failures so they can be recognized instead of re-described.

**expo-design-system**
- New `references/native-slop.md`: 20 named anti-patterns (The Web Modal, Everything's a Card, The Spinner Blink, ...) with the observable tell and the native replacement. Grep checks only for the 8 that grep cleanly, split into always-fix / review-each / advisory. The rest are marked screenshot-only.
- Top 5 tells inline in SKILL.md. audit.md points at the rest.
- Dynamic Type paragraph in Typography. `accessibilityLabel` added to the component contract and the audit. It had zero mentions in the corpus.

**expo-data-fetching**
- "Every screen has four states": loading is not empty, `ListEmptyComponent`, keep stale data on refetch, gate on hydration before redirecting.

**expo-native-ui**
- Keyboard: `keyboardShouldPersistTaps="always"` over a live keyboard, primary action never under the keyboard, pointer to the expo-animation recipe.
- Android floor: follow each platform's own HIG. SF Symbols do not render on Android, so every icon needs a Material source (`md` prop on NativeTabs, `@expo/vector-icons` Material families). icons.md gets an Android section.
- One-line pointer to the four-states rule, for screens built on mock data.

**expo-animation**, one line: `scale: 0.97` scoped to button-like pressables, rows highlight instead. This was contradicting tell #13, and the baseline eval's residual tells were Squish Reflex x3.

**Routing**: "looks AI-generated / not native" routes to expo-design-system in the overview and the description.

Plugin 1.12.4 → 1.13.0. The P3 branch also bumps to 1.13.0; whichever merges second re-bumps.


plugins/expo/skills/expo-design-system/references/native-slop.md
```diff
@@ -0,0 +1,74 @@
+# Native Slop: Named Anti-Pattern Tells
+
+The recurring failures of AI-generated React Native apps, each with a memorable name, an observable tell, and the native replacement. Before shipping a screen, scan it against this list by name: a tell that is present is a bug unless it is a deliberate, documented design decision of this app. The names double as review vocabulary - "that's a Floating Pill Tab Bar" is a complete review comment - and as a grading checklist for screenshots.
+
+## The 20 tells
+
+| # | Name | The tell (observable) | Native instead |
+|---|---|---|---|
+| 1 | **The Web Modal** | A centered rounded rectangle over a dimmed backdrop for picking, composing, or confirming - a web dialog teleported onto a phone | Bottom sheet (`presentation: 'formSheet'`, `@expo/ui` BottomSheet), action sheet, or anchored menu |
```
wodin: Should these be `**Web Modal**` rather than `**The Web Modal**`?
I would think the review would say something like "That's a Web Modal" rather than "That's The Web Modal"? But I suppose it could say "That's an instance of The Web Modal smell"?

davidmokos (approved):

PR #180, opened 2026-09-02
[skills] Add EAS Update skill

sjkim-expo:
## Why

Addresses [ENG-25468](https://linear.app/expo/issue/ENG-25468/add-agent-skill-for-eas-update).

Agents do not always choose the right path when configuring EAS Update.  They've shown behaviors like editing app configuration manually instead of running `eas update:configure`, and their knowledge / guidance on runtime compatibility and release-build activation can be incomplete or misleading.

Other examples and supporting context:
- A TestFlight user does not see an update after opening once
- An agent may say "reopen twice" without explaining why
- An agent may try to publish a native change as an OTA update
- Vague prompts like "push to prod" may lead an agent to attempt dangerous operations without first resolving the target

This adds a focused skill for configuring, publishing, testing, and debugging EAS Update without duplicating the full product documentation.

## How

I added an `eas-update` skill that starts by inspecting the project and uses the existing EAS CLI for configuration. It gives agents enough of the builds, channels, branches, updates, and runtime-version model to make compatibility decisions and explain why an installed build did or did not receive an update.

The skill covers bounded production publishing, download and apply lifecycle in release builds, and debugging instructions for project, channel, platform, runtime, and embedded configuration mismatches. It routes update health questions to `eas-update-insights` and native release work to `eas-app-stores`.

## Test Plan

Repository validation suite:

- `claude plugin validate .`
- `claude plugin validate ./plugins/expo`
- `bun scripts/check-skill-limits.ts`
- `bun scripts/check-overview-routing.ts`
- `bun scripts/check-plugin-version-bump.ts origin/main`
- `bun test scripts/check-plugin-version-bump.test.ts` (20 passed)
- JSON parsing and `git diff --check`

Fresh Claude sessions results against the local plugin:

| Scenario | Expected behavior | Observed |
| --- | --- | --- |
| Configure OTA updates | Inspect the project and use `expo install expo-updates` plus `eas update:configure` instead of inventing configuration | Loaded `eas-update`, installed the SDK-compatible package, and stopped before remote linking without partially reproducing `update:configure` by hand |
| TestFlight still shows old code | Explain background download and application on a later cold launch, then check channel, platform, and runtime compatibility | Three different phrasings loaded `eas-update` on `sonnet[1m]` and explained the up-to-two-cold-launch behavior |
| Publish a native dependency OTA | Reject the incompatible OTA path and require a new native build without changing runtime policy as a side effect | Loaded `eas-update`, refused to publish, required a new build, and preserved the existing runtime-version policy |
| Push a fix to production | Resolve the project, release type, channel, environment, platform, runtime, and authorization before publishing | Loaded the Expo router and stopped because the repository and production target were ambiguous |
| Check update health | Route crash, adoption, and rollout questions away from the general publishing workflow | Did not select `eas-update` and identified `eas-update-insights` as the appropriate workflow |
| Release changed iOS entitlements | Recognize native configuration and use the store-release workflow | Loaded `eas-app-stores` and required a new native build |
| Switch one QA installation to preview | Use per-installation channel surfing without globally remapping production | Loaded `eas-update`, rejected `eas channel:edit`, and used the device-local request-header override |

### DIFF
--- a/plugins/expo/skills/eas-update/SKILL.md
+++ b/plugins/expo/skills/eas-update/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: eas-update
-description: "EAS service (paid). Configure and use EAS Update for over-the-air JavaScript and asset updates with expo-updates and EAS CLI. Use when setting up OTA updates, running eas update:configure or eas update, publishing to preview/staging/production channels, explaining branches/channels/runtime versions, testing an update in a development or release/TestFlight build, or debugging why a build did not receive an update. Not for update health metrics; use eas-update-insights for adoption, crashes, and rollout monitoring."
+description: "EAS service (paid). Configure and use EAS Update for over-the-air JavaScript and asset updates with expo-updates and EAS CLI. Use when setting up OTA updates, running eas update:configure or eas update, publishing to preview/staging/production channels, explaining branches/channels/runtime versions, testing updates, or debugging why an installed build still shows old code. Load for TestFlight, preview, or production updates that do not appear, including questions about cold launches or reopening the app. Not for update health metrics; use eas-update-insights for adoption, crashes, and rollout monitoring."
 version: 1.0.0
 license: MIT
 allowed-tools: "Bash(npx expo *), Bash(npx *eas-cli@*), Bash(eas *)"
@@ -58,6 +58,8 @@ Use this model to make decisions, but explain only the concepts needed for the u
 Use an update for changes to JavaScript, styling, and bundled assets that the installed native runtime already supports.
 
 Create a new native build when a change adds or modifies native code or native configuration, including most native-library additions and SDK upgrades. Do not work around a runtime mismatch or imply that publishing can add native capabilities to an existing build. See https://docs.expo.dev/eas-update/runtime-versions/.
+
+Do not change the project's runtime-version policy as an incidental fix. Explain how the current policy affects compatibility; treat changing it as a separate decision because it changes which installed builds can receive future updates.
 
 ## Publish deliberately
 
--- a/plugins/expo/skills/eas-update/SKILL.md
+++ b/plugins/expo/skills/eas-update/SKILL.md
@@ -30,6 +30,8 @@ npx eas-cli@latest update:configure
 
 Use `eas update:configure` rather than manually inventing `updates.url`, `runtimeVersion`, native metadata, or build-profile channels. The command understands EAS project linking, Continuous Native Generation, and projects with committed native directories. Review and explain its resulting diff.
 
+If the command cannot proceed because the project is not linked or the user has not authorized the required remote operation, stop after any independently valid package installation and explain what remains. Do not partially reproduce `update:configure` by adding a runtime-version policy, config plugin, update URL, or channels by hand.
+
 For dynamic app config, non-EAS builds, or a command that cannot complete automatically, follow the current setup documentation instead of guessing: https://docs.expo.dev/eas-update/getting-started/.
 
 ## Keep the model straight
--- a/plugins/expo/skills/eas-update/SKILL.md
+++ b/plugins/expo/skills/eas-update/SKILL.md
@@ -14,7 +14,14 @@ Use EAS Update to deliver compatible JavaScript, styling, and asset changes to i
 
 ## Start with the supported configuration path
 
-Before changing anything, inspect `package.json`, the Expo app config, `eas.json` if present, and whether `ios/` or `android/` are committed. Detect the Expo SDK version and preserve existing dynamic or platform-specific configuration.
+Before changing anything, inspect `package.json`, the Expo app config, `eas.json` if present, and whether `ios/` or `android/` are tracked. Use what you find when reviewing the CLI's changes:
+
+- Preserve existing dynamic or platform-specific app configuration.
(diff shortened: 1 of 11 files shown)


plugins/expo/skills/eas-update/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Advanced and adjacent workflows
+
+- **Channel surfing:** an individual release build can override its `expo-channel-name` request header to request another compatible channel. This differs from changing the server-side channel-to-branch mapping. Follow https://docs.expo.dev/eas-update/channel-surfing/ and preserve its access-control, persistence, recovery, and compatibility constraints.
+- **Update health:** load `eas-update-insights` for adoption, launch failures, crash rate, payload size, and rollout monitoring after publishing.
+- **Store releases:** load `eas-app-stores` when native changes require a new TestFlight, App Store, or Play Store build.
+
+## Official references
+
+- Setup: https://docs.expo.dev/eas-update/getting-started/
```
davidmokos: I think we can reference the direct markdown links, i.e. https://docs.expo.dev/eas-update/getting-started.md

plugins/expo/skills/eas-update/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+
+# EAS Update
+
+> **EAS service - costs apply.** EAS Update is available on the Free plan; publishing and delivery use update, bandwidth, and storage allowances, with higher limits on paid plans. See https://expo.dev/pricing.
+
+Use EAS Update to deliver compatible JavaScript, styling, and asset changes to installed apps without submitting a new native binary. Native-code changes still require a new build.
+
+## Start with the supported configuration path
+
+Before changing anything, inspect `package.json`, the Expo app config, `eas.json` if present, and whether `ios/` or `android/` are committed. Detect the Expo SDK version and preserve existing dynamic or platform-specific configuration.
```
davidmokos: I'm confused by this instruction. What should I do if eas.json is or isn't present, or when `/ios` or `/android` folders are or aren't committed? It only tells the agent to check but not what to do with the information.

davidmokos (approved):
Small nits inline. Feel free to merge, the work on the evals can be done async - having any update skill is better than none anyway :)

Kudo:
> Agents do not always choose the right path when configuring an EAS update. They've shown behaviors like editing app configuration manually instead of running eas update:configure, and their knowledge / guidance on runtime compatibility and release-build activation can be incomplete or misleading.

i'm wondering what problems the agent struggled on. did it fetch any resource from https://docs.expo.dev? good to share some prompts and results with and without the skill as comparison.

i didn't mean to push back for having the skill. it makes sense to have the skill. i just want to know what's missing on our doc. otherwise, it feels like agents don't get any help from our doc. then we should convert all our docs to skills. that doesn't feel correct.

plugins/expo/skills/eas-update/SKILL.md
sjkim-expo: Yeah good point. The goal was to help the agent review what `update:configure` changed, but the instruction tells it to read project state without explaining how it affects the workflow. 

Will change this to have more explicit instructions:
- reveiw channel additions when `eas.json` exists
- do not create one manually when it is absent (since we've seen agents do this before)
- distinguish tracked native projects from CNG when reviewing the configuration generated by the CLI

sjkim-expo:
> > Agents do not always choose the right path when configuring an EAS update. They've shown behaviors like editing app configuration manually instead of running eas update:configure, and their knowledge / guidance on runtime compatibility and release-build activation can be incomplete or misleading.
> 
> i'm wondering what problems the agent struggled on. did it fetch any resource from https://docs.expo.dev? good to share some prompts and results with and without the skill as comparison.
> 
> i didn't mean to push back for having the skill. it makes sense to have the skill. i just want to know what's missing on our doc. otherwise, it feels like agents don't get any help from our doc. then we should convert all our docs to skills. that doesn't feel correct.

@Kudo Current tests show that the skill produces the intended behavior, but don't isolate whether the underlying issue is the docs, retrieving the docs or applying them correctly.

In the original example with needing to launch TestFlight twice (see Slack thread with Joe: https://exponent-internal.slack.com/archives/C0B96B57KGQ/p1784859335758529), Codex did cite the relevant docs, but the user still had to ask why two launches were neeed. 

If we want, we can try to run the same configuration and troubleshooting promptos with and wo the skill, and record which docs the agent fetches / the actions it takes in both scenarios. 

FWIW, @davidmokos and I spoke earlier and think it might just be worth it to release the skill first, and see how customers interact with it since it's pretty low-stakes.

plugins/expo/skills/eas-update/SKILL.md
sjkim-expo: Updated these to use the direct `.md` links throughout the skill.

plugins/expo/skills/eas-update/SKILL.md
sjkim-expo: Updated this so each check now affects how the agent reviews the CLI's changes, including the behavior with or without `eas.json` and with tracked native projects versus CNG.

Kudo:
> If we want, we can try to run the same configuration and troubleshooting promptos with and wo the skill, and record which docs the agent fetches / the actions it takes in both scenarios.

yes if you have some chance, that's something i'm curious and like to know what we can improve in our docs.


btw the skill itself looks good. let's merge it!

Kudo (approved):

PR #183, opened 2026-09-06
Extend EAS skills to support native Swift iOS apps

zvadaadam:
Adds native Swift iOS support to the EAS skills, based on testing a SwiftUI app.

- Updates skill routing and adds native setup guidance so Swift apps don't get Expo scaffolding.
- Corrects CLI commands and clarifies TestFlight processing and App Store review.
- Documents targeted archive and icon checks for the stale build number and icon alpha failures found during testing.
- Updates skill listings and bumps all plugin manifests to 1.13.1.

Validation: checked existing native artifacts; 20 tests and repository checks pass.


zvadaadam:
This is the first step in improving expo/skills to work better with native apps.

My main focus was on getting EAS to work within a native app (Swift app).

I successfully built my native app and pushed it to TestFlight using these instructions.

davidmokos (approved):
do we need the python scripts? Can the same be achieved with a combination of eas cli and launch?

zvadaadam:
> do we need the python scripts? Can the same be achieved with a combination of eas cli and launch?

good judgment, u r right.

PR #184, opened 2026-09-07
Improve brownfield guidance and add runnable iOS fixtures

zvadaadam:
This is the next step in improving Expo skills for native apps, following #183. It focuses on adding Expo and React Native screens to an existing Swift app while keeping its native navigation.

- Adds examples for passing data between Swift and React Native, returning a result, and closing the screen.
- Fixes setup instructions for both building React Native with the app and packaging it as a library.
- Updates the guidance for SDK 57 and adds small test apps under `tests/fixtures/expo-brownfield` so we can try changes and check future SDK releases.

Tested both approaches in Debug and Release on iOS simulators with SDK 55 and SDK 57. The SDK 57 runs used Xcode 26.6 and iOS 26.5, including Release with Metro stopped. CI passes.


davidmokos (approved):
before merging I think we should test with SDK 57 and the latest xcode

zvadaadam:
> before merging I think we should test with SDK 57 and the latest xcode

run a test a it works, also dropping some irrelevant wording about versions on which this was tested. i don't that should be part of skills.

davidmokos (commented):
we should remove the tests fixtures

PR #187, opened 2026-09-09
[eas-simulator] cover current CLI flags

szdziedzic:
## Summary

- document every public EAS Simulator flag and command
- clarify that idle timers reset only for agent-device and argent activity
- add Appium and web-preview-only guidance and correct JSON/dotenv behavior
- guard against orphaning an existing session when starting another

## Test plan

- bun scripts/check-skill-limits.ts
- bun scripts/check-overview-routing.ts
- bun scripts/check-plugin-version-bump.ts origin/main
- claude plugin validate ./plugins/expo
- git diff --check
- compare command coverage with eas-cli 23.2.0 help


tchayen (approved):

plugins/expo/skills/eas-simulator/references/cli-reference.md
```diff
@@ -0,0 +1,51 @@
+# EAS Simulator CLI reference
```
gwdp, line 1: I would be a little hesitant about needing to keep this in sync with whatever the CLI changes. What do you think about leveraging CLI --help flag or something similar?

plugins/expo/skills/eas-simulator/references/run-your-app.md
```diff
@@ -1,18 +1,20 @@
 # Running your app on the remote sim — tested sequences
 
-The remote sim boots blank. You install a **simulator-targeted** build onto the session, then open it. Pick a mode from `SKILL.md`. (Sequences validated against eas-cli 20.3.x + agent-device 0.17.x in mid-2026; the commands are experimental — if one fails, re-check `<cmd> --help`.)
+The remote sim boots blank. You install a **simulator-targeted** build onto the session, then open it. Pick a mode from `SKILL.md`. (Sequences validated against eas-cli 20.3.x + agent-device 0.17.x in mid-2026; the CLI surface was rechecked against eas-cli 23.2.0. These commands are experimental — if one fails, re-check `<cmd> --help`.)
```
gwdp, line 3: `the CLI surface was rechecked against eas-cli 23.2.0.` might not be needed since we already say `Sequences validated against eas-cli 20.3.x` in the same paragraph?

plugins/expo/skills/eas-simulator/references/run-your-app.md
```diff
@@ -1,18 +1,20 @@
...
 In all modes, the session is started the same way and driven through `npx --yes eas-cli@latest simulator:exec`. Replace `dev.example.app` with the app's iOS `bundleIdentifier` (from `app.json` → `ios.bundleIdentifier`), and run from the project directory.
 
-> These sequences are **iOS**. For **Android**: build via `npx --yes eas-cli@latest build --platform android` (or local Gradle), `install` the `.apk` instead of an `.app`, skip `pod install`, and note there's **no `webPreviewUrl`** (Android is agent-driven / screenshot-only).
+> These sequences are **iOS**. For **Android**: build via `npx --yes eas-cli@latest build --platform android` (or local Gradle), `install` the `.apk` instead of an `.app`, and skip `pod install`. Current simulator session types include a web preview, though Android support is still in development and may lack iOS parity.
 
 ## Starting a session (shared by all modes)
 
 ```bash
-# Reset the dotenv first so the new session id isn't masked by an "Overwriting previous session" warning.
+# If the dotenv names a session, inspect it first with simulator:get --json. Stop it if IN_PROGRESS;
```
gwdp, line 12: `Stop it if IN_PROGRESS` might conflict with users trying to keep multiple active sessions?

gwdp (changes requested):
Nice improvements! Unsure about the cli-reference.md, mainly due to the maintenance burden it could create. If we could figure out a different way that does require us to update the skill every time we add something in the CLI, imo it would be preferable. Let me know if you have different thoughts 🙇‍♂️

plugins/expo/skills/eas-simulator/references/cli-reference.md
szdziedzic: Agreed. I removed the static cli-reference.md cache and updated SKILL.md to query the installed eas-cli subcommand help for the complete current flag set, while retaining non-obvious behavioral guidance such as idle-timer resets. The current help already includes egress flags that were absent from the snapshot, which confirms this avoids drift.

plugins/expo/skills/eas-simulator/references/run-your-app.md
szdziedzic: Removed the eas-cli 23.2.0 recheck note and now direct agents to the relevant subcommand help before using non-default flags. I also removed the other fixed 23.2.0 verification note from controllers.md.

plugins/expo/skills/eas-simulator/references/run-your-app.md
szdziedzic: Updated this flow plus the matching core-loop and troubleshooting guidance. Agents now reuse a session when it belongs to the current run, stop it only when it is in scope and no longer needed, and treat IN_PROGRESS as potentially intentional concurrency by preserving its id/config before resetting the dotenv.

gwdp (approved):

PR #192, opened 2026-09-10
[expo-ui] Document RNHostView matchContents sizing

intergalacticspacehighway:
# Why

`RNHostView` with `matchContents` measures its hosted child itself, from that child's own point-valued `minWidth` / `maxWidth` / `minHeight` / `maxHeight`. The width the native parent offers never reaches the child. So an unbounded `<Text>` measures at its full single-line width and runs off the screen instead of wrapping.

Nothing in the `expo-ui` skill said this, so agents write the overflowing version.

Source: https://github.com/expo/expo/pull/49483

# How

- `references/swift-ui.md` and `references/jetpack-compose.md`: new `RNHostView` sizing section with a ✅ / ❌ pair, plus three rules the PR establishes — only the first child is measured and laid out, `alignSelf: 'auto' | 'stretch'` is forced to `flex-start`, and `onLayout` reports the size the host settled on.
- `references/universal.md`: an `RNHostView` row in the components table and a compact note pointing at the platform references.
- Bumped all four plugin manifests to 1.13.2.

The guidance is scoped to `matchContents`. Without it the host fills its native parent and text wraps on its own, which the new section says explicitly.

# Test Plan

```
bun scripts/check-skill-limits.ts             # passes
bun scripts/check-plugin-version-bump.ts origin/main   # passes, 1.13.1 → 1.13.2
bun scripts/check-overview-routing.ts         # passes
claude plugin validate .                      # passes
claude plugin validate ./plugins/expo         # passes
```

Behaviour claims are taken from the merged expo/expo PR — `ExpoViewShadowNode::hostedContentConstraints` for which style bounds the measure reads, `ExpoViewComponentDescriptor::adopt` for the `alignSelf` override, and the `ExpoUIRNHostViewSize` test suite (`honours the hosted view's own maxWidth and minHeight`) for the wrapping case.

https://claude.ai/code/session_01Np8hnE1biHAoRiAcx42vqZ

### DIFF
--- a/plugins/expo/.claude-plugin/plugin.json
+++ b/plugins/expo/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "expo",
-  "version": "1.13.1",
+  "version": "1.13.2",
   "description": "Official Expo skills for building, deploying, upgrading, and debugging Expo apps",
   "author": {
     "name": "Expo Team",
--- a/plugins/expo/.codex-plugin/plugin.json
+++ b/plugins/expo/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "expo",
-  "version": "1.13.1",
+  "version": "1.13.2",
   "description": "Official Expo skills for building, deploying, upgrading, and debugging Expo and React Native apps.",
   "author": {
     "name": "Expo Team",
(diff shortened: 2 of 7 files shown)


Kudo (approved):
the content is pretty much the same. maybe we could move it to dedicated reference md and link it

some idea

```md
## `RNHostView` sizing with `matchContents`

If you want `RNHostView` with `matchContents`, check [RNHostView.md](./RNHostView.md) to do it correctly.
```


i like the
```
// ❌ Measures as one line and overflows
// ✅ Wraps at 280 points
```

and the universal component doesn't have it. that's why i was thinking to split it to a reference file

PR #196, opened 2026-09-14
[repo] remove root tests directory

davidmokos:
Removes the root `tests/` directory added in #184 and its documentation links. Bumps the plugin version to 1.13.5 for the skill reference change.

Plugin validation, skill limits, overview routing, the plugin version check, and all 20 version-check tests pass.


Kudo (approved):
do you have a recommended way where to put example for skills?

